PR #67, opened 2020-10-15
Adds api call to coingecko and renders DPI market data to homepage

kibagateaux:
Changes:
- Adds `MarketDataProvider` which pulls price, marketcap, and volume data in the last month DPI. I think each array item is each hour's 24 hour rolling data, the last timestamped data matches up to the 24hour data on Coingecko's site so it seemed right and I didn't look into it further.
- Adds 24 hour volume and current marketcap of DPI to home page (see screenshots)
- Adds graph of price movements for last month for DPI to home page (see screenshots)
- Adds rechart.js library. I don't think we are going to be doing crazy graphs and this one is pretty simple and popular. Open to whatever suggestions for a different library.
- Moves Uniswap liquidity and $INDEX price to farming page.
- Rearranges content on homepage for a more marketing type flow

TODOS:
- Move coingecko marketdata API call to an util function so we can change the token/timeframe of query
- Price graph UI updates based on feedback
- Add UI to change between price and marketcap on graph
- Fix tooltips on graph, were acting buggy for me on Brave browser
- Tests? lol

Homepage:
Screen Shot 2020-10-15 at 2 03 02 PM

...otherstuff...

Screen Shot 2020-10-15 at 2 03 20 PM

Farming page:
Screen Shot 2020-10-14 at 6 59 14 PM


controtie (commented):
Awesome, this looks generally good! Some general feedback (knowing this is still WIP)
* Set your editor from tabs -> 2 spaces for this repo
  * I'll work on getting an auto linter/formatter set up on commit too
* Lean to use more descriptive variable names
  * Where there's a tradeoff between a clearer variable name but longer one and a less clear but shorter one, take the longer variable name
  * Examples: [MarketDataChart.tsx](https://github.com/SetProtocol/index-ui/pull/67/files#diff-0ad9509f82027e30ccc1e0d2f32a793c80f1ca3318e23734cb86900865158cdaR34-R35), [MarketDataProvider.ts](https://github.com/SetProtocol/index-ui/pull/67/files#diff-27a7a54067907df8366d84cebce0a9578f176b54407f262fa8e7b72b93278c5cR8)

kibagateaux:
@controtie 
Simple changes:
- Fixed code formating to spaces
- added Tooltip to price chart
- added latest price and monthly change to DPI data on homepage
- Updated variable names, still feel like they aren't right but let me know what you guys think.

**Token data provider pattern:**
We discussed context/provider in discord and I did some more research on it but I'll defer to you on this since this is my first time using them and don't know the best practices. I don't think putting all the token data in the same provider is right because if we only need one token we have to make api requests for all of them. :
- Moved `DpiTokenMarketDataProvider` to Home page since we only need it there. If we are going to be adding all the underlying tokens we don't want 12 api calls when we might not even be using that data on every page (e.g. /farms)
- I wanted to make `DpiTokenMarketDataProvider` a general provider pattern to use with any tokens but since the React hooks use the Context instead of Provider to inject data (e.g. `useDpiTokenMarketData()` uses `DpiTokenMarketDataContext` instead of `DpiTokenMarketDataProvider` which doesn't make sense to me.) I'm not sure how to use a general Context pattern that can be applied to multiple token data Providers (maybe they aren't supposed to?).
- `DpiTokenMarketDataProvider` should actually be used to get any token data but above was an issue. Ideally it would take `coingeckoId` and `daysOfHistory` so when we make requests for all the underlying tokens we can update a couple state variables and this will update all the API requests we need. I envisioned something like this:
```
<DpiIndexDetailsPage>
  <TokenMarketDataProvider coingeckId='defipulse-index' daysOfHistory={someUserInput}>
     <TokenMarketDataProvider coingeckoId='ren' daysOfHistory={someUserInput}>
         {pageContentComponents}
     </>
  </>
</DpiIndexDetailsPage>
```
This way we can easily compose single or multiple tokens together by just mapping over the coingecko ids e.g. 
```
DpiUnderlyingTokenDataProviders = ['ren', 'yearn-finance',...].map(createTokenDataProviders)
<DpiIndexDetailsPage>
  <DpiUnderlyingTokenDataProviders>
     {pageContentComponents}
  </>
</>
```
Main issue with this is we can't use Promise.all for the API requests but we can also make `coingeckoId` prop take an array if we wanted. It also requires some input to give the Provider functionality which I don't know if thats a best practice.

I was trying to make it easy for you when you do whatever for the underlying index tokens but we can totally disregard this and make DPI stuff simple and hardcoded if you want. Most of the functionality is abstracted now with the coingecko util function so incase this was wrong, we can revert that decision easily just by changing variables to be hardcoded.

I get on my flight to Korea in like 30 minutes so I won't be able to respond for maybe 24 hours.

src/contexts/DpiTokenMarketData/DpiTokenMarketDataProvider.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState, useEffect } from 'react'
+import MarketDataContext from './DpiTokenMarketDataContext'
+import { coingeckoDpiId } from 'constants/coingeckoIds'
+import { fetchHistoricalTokenMarketData } from 'utils/coingeckoApi'
+
+interface HistoricalMarketDataProvider {
+  // coingeckoId: string
+	daysOfHistory: number
```
controtie, line 8: I'm surprised this didn't get picked up by the auto linter? Let me know if it isn't working for you. You should be able to make a small change to this file, commit it and it should fix this kind of stuff.

src/utils/coingeckoApi.ts
```diff
@@ -0,0 +1,31 @@
+const baseURL = 'https://api.coingecko.com/api/v3'
+
+export const fetchHistoricalTokenMarketData = (
+	id: string,
+	startTime: number | string,
+	endTime: number | string,
+	baseCurrency:string = 'usd'
```
controtie, line 7: You can omit `string` here since you are assigning 'usd' as default.

src/views/Farm/Farm.tsx
```diff
@@ -57,6 +58,8 @@ const Farm: React.FC = () => {
         title="Farm"
       />
       <Spacer size="md" />
+			<Treasury />
+			<Spacer />
```
controtie, line 62: Spacing

src/views/Home/components/MarketData.tsx
```diff
@@ -0,0 +1,87 @@
...
+import useDpiTokenMarketData from 'hooks/useDpiTokenMarketData';
+
+const MarketData: React.FC = () => {
+  const {
+    latestVolume,
+    latestMarketCap,
+    latestPrice,
+    prices
+  } = useDpiTokenMarketData()
+  const priceAtEpochStart = (prices || [[0, 1]])[0][1]
```
controtie, line 19: The second part of this expression is sort of hard to understand.
Our repo has access to [optional chaining which is pretty cool](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

It looks sort of weird with dynamic array indices but it would work like this here:
`const priceAtEpochStart = (prices?.[0]?.[1] || 1)`

src/contexts/DpiTokenMarketData/DpiTokenMarketDataProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const startTime = endTime - (86400 * daysOfHistory);
+    fetchHistoricalTokenMarketData(coingeckoDpiId, startTime, endTime)
+      .then((response: any) => {
+        setDpiMarketData(response)
+      })
+      .catch((error: any) => console.log(error));
+  }, [])
+
+  const getLatest = (arr?: number[][]) =>
+    arr && arr.length ? arr[arr.length - 1][1] : 0
```
controtie, line 27: Label these a bit more clearly. `arr` could be `tokenMarketData`? and `getLatest` could be `selectLatestMarketData`?

Also consider using optional chaining here

controtie (changes requested):
Ok, good progress. And the tool tip looks nice 👌

Let's move the `DpiTokenProvider` back to the top level `App.tsx` for now. I'm not 100% sure about this, but I don't think there's actually very much efficiency gains from placing it in the `Home` page. Only components which invoke `useDpiMarketData` should trigger re-renders when `DpiMarketDataProvider` changes.

As far as having a token generalized interface, let's not worry too much about that now. Depending on how we want to present the data, we may just be able to fetch 1 token at a time depending on what the user wants to see as a benchmark.

I would say this PR is like 90% there. Let's focus on getting this PR under your belt first and then we can move on to some of these more detailed conversations 🙂

src/views/Home/components/MarketData.tsx
kibagateaux: ahh ok I was trying that, didn't realize you had to add the . after ? so I figured it wasn't possible with arrays. Thanks.

src/contexts/DpiTokenMarketData/DpiTokenMarketDataProvider.tsx
kibagateaux: For function name I'l definitely change,  called it `getLatest()` because it's basically just a double `last()`.
For the function body optional chaining doesn't make the code that much shorter and for me, my original code is easier to read but I've never used optional chaining before so that could just be me
`marketData?.[marketData.length -1]?.[1] || 0` vs `arr && arr.length ? arr[arr.length - 1][1] : 0`

controtie: Ok let's keep it in your original style then, just with updated variable names.

kibagateaux: Oh it's actually uncomfortably long once the variable name is changed, I'll just use the optional chaining lol

kibagateaux:
Thanks for feedback, everything should be good to go now

src/utils/coingeckoApi.ts
```diff
@@ -0,0 +1,31 @@
...
+			 * Daily data will be used for duration above 90 days.
+			*/
+			const {
+				prices,
+				market_caps: marketcaps,
+				total_volumes: volumes,
+			} = response
+			return { prices, marketcaps, volumes }
+		})
+		.catch(error => console.log(error));
```
controtie, line 30: Tabs -> spaces in this file

controtie (changes requested):
One more minor code styling issue. Looks good though 👌

src/utils/coingeckoApi.ts
kibagateaux: oops. done.

controtie (approved):
Looks good 👌

PR #75, opened 2020-10-28
Dpi product

kibagateaux:
Don't have permissions on repo so can't push directly to branch :/
Adds:
* chart price/percentage updates when scrolling over chart
* all token stat components from design. Only missing the About text section, will throw that in after review
* adds /dpi to nav menu. Doesn't change links on homepage, v2 spec says "buy" buttons point there but we don't have buying yet so left as is.
* Placeholder "buy" button on DPI page. Have some design/product questions before I finish up the layout aspect of it.

Missing some design assets to finish off the spec:
* Mono font for chart prices
* hsl/hex/rgba codes for red/green prices
* Icon assets for buy/sell button

Getting a warning on chart price update function: `Cannot update component 'MarketData' from other component 'Tooltip'`
It only occurs once you start scrolling on the chart, not on initial render, and even if I did checks on the data to prevent potentially unnecessary updates.  It's only a warning but seemed to have a noticeable impact on chart rendering performance even when I ran a built version of the site so I'll do some digging. Chose to use the `readTooltipData` the other functions exposed by the chart e.g. onMouseMove don't actually tell you the actual x,y values where the mouse is but I might find a better solution while I'm fixing the issue.

### DIFF
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -40,15 +40,13 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
 
   const renderTooltip = (props: any) => {
     const tooltipData = props.payload?.[0]
-    readTooltipData && readTooltipData(tooltipData)
+
+    if (readTooltipData && tooltipData) readTooltipData(tooltipData)
+
     if (!showTooltip) return null
 
     const [label, value] = formatToolTip(tooltipData)
     return <FancyValue icon={icon} label={label} value={value} />
-  }
-
-  const onMouseMove = (props: any) => {
-    console.log('chart price hover', props)
   }
 
   const minY = Math.min(...(data || []).map<number>(({ y }) => y))
@@ -64,7 +62,6 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
             dot={false}
             stroke={'url(#gradient)'}
             strokeWidth={2}
-            onMouseMove={onMouseMove}
           />
           <YAxis
             stroke={theme.colors.grey[500]}
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -18,6 +18,7 @@ interface SimplePriceChartProps {
     x: string | number
     y: number
   }[]
+  readTooltipData?: Function
 }
 
 const MarketDataChart: React.FC<SimplePriceChartProps> = ({
@@ -25,6 +26,7 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   showTooltip,
   icon,
   data,
+  readTooltipData,
 }) => {
   const theme = useTheme()
   const formatFloats = (n: number) => parseFloat(numeral(n).format('0.00a'))
@@ -37,10 +39,16 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   }
 
   const renderTooltip = (props: any) => {
+    const tooltipData = props.payload?.[0]
+    readTooltipData && readTooltipData(tooltipData)
     if (!showTooltip) return null
 
-    const [label, value] = formatToolTip(props.payload?.[0])
+    const [label, value] = formatToolTip(tooltipData)
     return <FancyValue icon={icon} label={label} value={value} />
+  }
+
+  const onMouseMove = (props: any) => {
+    console.log('chart price hover', props)
   }
 
   const minY = Math.min(...(data || []).map<number>(({ y }) => y))
@@ -56,6 +64,7 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
             dot={false}
             stroke={'url(#gradient)'}
             strokeWidth={2}
+            onMouseMove={onMouseMove}
           />
           <YAxis
             stroke={theme.colors.grey[500]}
--- a/src/views/DPI/components/DpiHoldings.tsx
+++ b/src/views/DPI/components/DpiHoldings.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React from 'react'
 import styled from 'styled-components'
 import numeral from 'numeral'
 
@@ -17,7 +17,6 @@ const DpiHoldings: React.FC = () => {
         ${numeral((latestPrice || 0) * Number(dpiBalance)).format('0.00a')}
       </BaseCurrencyDpiValuation>
       <DpiTokenHoldings>
-        {' '}
         {numeral(dpiBalance).format('0.000a')} DPI
       </DpiTokenHoldings>
     </InfoSection>
--- a/src/views/DPI/components/DpiHoldings.tsx
+++ b/src/views/DPI/components/DpiHoldings.tsx
@@ -1,0 +1,39 @@
+import React, { useState, useEffect } from 'react'
+import styled from 'styled-components'
+import numeral from 'numeral'
+
+import InfoSection from './InfoSection'
+
+import useBalances from 'hooks/useBalances'
+import useDpiTokenMarketData from 'hooks/useDpiTokenMarketData'
+
+const DpiHoldings: React.FC = () => {
+  const { dpiBalance } = useBalances()
+  const { latestPrice } = useDpiTokenMarketData()
+  console.log('DPI assets', Number(dpiBalance), latestPrice)
+  return (
+    <InfoSection title='My Assets'>
+      <BaseCurrencyDpiValuation>
+        ${numeral((latestPrice || 0) * Number(dpiBalance)).format('0.00a')}
+      </BaseCurrencyDpiValuation>
+      <DpiTokenHoldings>
+        {' '}
+        {numeral(dpiBalance).format('0.000a')} DPI
+      </DpiTokenHoldings>
+    </InfoSection>
+  )
+}
+
+const BaseCurrencyDpiValuation = styled.h3`
+  display: inline-block;
+  margin: 0 ${({ theme }) => theme.spacing[4]}px 0 0;
+  font-size: 28px;
+`
+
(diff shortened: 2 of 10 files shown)


src/views/DPI/components/DpiPriceChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [latestTime, latestPrice] = prices?.[prices.length - 1] || [0, 0]
+  // Coingecko API returns each item as hourly data if < 90 days and daily data > 90 days
+  // So if time between start and end time < 90 days, 1 day of data is 24 array items.
+  const dataCadenceFor24Hours = latestTime - startTime < 7776000000 ? 24 : 1
+  const priceChangeIntervals = [
+    ['1 Day', 1],
+    ['1 Month', 30],
+    ['3 Months', 90],
+    ['1 Year', 365],
+  ]
```
controtie, line 21: 👍

src/views/DPI/components/DpiHoldings.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import InfoSection from './InfoSection'
+
+import useBalances from 'hooks/useBalances'
+import useDpiTokenMarketData from 'hooks/useDpiTokenMarketData'
+
+const DpiHoldings: React.FC = () => {
+  const { dpiBalance } = useBalances()
+  const { latestPrice } = useDpiTokenMarketData()
+  console.log('DPI assets', Number(dpiBalance), latestPrice)
```
controtie, line 13: lint

src/views/DPI/components/DpiPriceChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import useDpiTokenMarketData from 'hooks/useDpiTokenMarketData'
+
+const DpiPriceChanges: React.FC = () => {
+  const { prices } = useDpiTokenMarketData()
+  const [startTime] = prices?.[0] || [0]
+  const [latestTime, latestPrice] = prices?.[prices.length - 1] || [0, 0]
+  // Coingecko API returns each item as hourly data if < 90 days and daily data > 90 days
+  // So if time between start and end time < 90 days, 1 day of data is 24 array items.
+  const dataCadenceFor24Hours = latestTime - startTime < 7776000000 ? 24 : 1
```
controtie, line 15: Renaming `7776000000` to a variable named `ninetyDays` would be helpful for me

src/views/DPI/components/DpiPriceChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+    ['1 Day', 1],
+    ['1 Month', 30],
+    ['3 Months', 90],
+    ['1 Year', 365],
+  ]
+
+  const calculatePriceDiffs = (daysOfComparison: number) => {
+    // placeholder if no data available for timerange
+    if (!prices || prices.length < daysOfComparison * dataCadenceFor24Hours)
+      return <div>---</div>
```
controtie, line 26: lets wrap these in braces unless it's strictly a 1 liner (e.g. fits on line 25 here) and is below 120 characters

src/views/DPI/components/DpiPriceChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+  // So if time between start and end time < 90 days, 1 day of data is 24 array items.
+  const dataCadenceFor24Hours = latestTime - startTime < 7776000000 ? 24 : 1
+  const priceChangeIntervals = [
+    ['1 Day', 1],
+    ['1 Month', 30],
+    ['3 Months', 90],
+    ['1 Year', 365],
+  ]
+
+  const calculatePriceDiffs = (daysOfComparison: number) => {
```
controtie, line 23: This function is doing two things: calculating the price difference & returning a JSX component.
Maybe separating the price change calculation into a util would help make it a bit easier to read.
Also just renaming the function to `renderAssetPriceChange` could help.

src/views/DPI/components/DpiPriceChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const priceDiff = ((latestPrice - startPrice) / startPrice) * 100
+    const StyledPrice =
+      priceDiff < 0 ? StyledNegativePrice : StyledPositivePrice
+    return (
+      <StyledPrice>{numeral(priceDiff).format('+0.00a') + '%'}</StyledPrice>
+    )
+  }
+
+  const renderPriceChanges = (config: any) => {
+    const [dateString, daysOfComparison] = config // how to destructure in params without tsx bitching?
```
controtie, line 39: Maybe adding a type for config might help with the linter error?

src/views/DPI/components/DpiPriceChanges.tsx
```diff
@@ -0,0 +1,78 @@
...
+    return (
+      <StyledPriceChange key={daysOfComparison}>
+        <StyledDateString>{dateString}</StyledDateString>
+        {calculatePriceDiffs(daysOfComparison)}
+      </StyledPriceChange>
+    )
```
controtie, line 45: Yeah, it looks like the positive and negative style price could be moved into this section (unless I'm missing something)

src/views/DPI/components/InfoSection.tsx
```diff
@@ -0,0 +1,30 @@
...
+      <SectionTitle>{title}</SectionTitle>
+      {children}
+    </SectionContainer>
+  )
+}
+
+const SectionContainer = styled.div`
+  padding: ${({ theme }) => theme.spacing[6]}px 0;
+
+  :not(:last-child) {
```
controtie, line 20: neat

controtie (commented):
Looks generally good. I'll ping @inje for the font and color codes.
I think the buy/sell assets you're using now actually looks great, so maybe just fitting them into the specced lay-out and we're good to go.

controtie:
Here are the red and green colors
Green #1FC36A
Red #DC3F4A

Can add the font in another PR. It's Robot Mono

src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -37,9 +39,13 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   }
 
   const renderTooltip = (props: any) => {
+    const tooltipData = props.payload?.[0]
+
+    if (readTooltipData && tooltipData) readTooltipData(tooltipData)
```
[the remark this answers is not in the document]
controtie: Maybe add a debouncer to it as well

src/views/DPI/components/DpiPriceChanges.tsx
kibagateaux: Tried that but then it complained on the .map line about not declaring all the other parameters to the function but for some reason doesn't with this setup. I'll just leave it as is

src/views/DPI/components/DpiPriceChanges.tsx
kibagateaux: I think I had that at first but then the logic for checking for the placeholder '---' before doing numeric operation seemed less clean than the early return in the `calculatePriceDiffs` function. I thought about making a single component called `StyledPrice` which took the % diff as a prop and calculated the color in CSS 
 instead of JS but that didn't seem right either. e.g.
```
const priceDiff = calcDiff(days)

const StyledPrice= styled.div'${({priceDiff}) =>
  "color: ${!Number(priceDiff) /* placeholder */ ? 'white'  : priceDiff < 0 ? 'red' : 'green' "}'

<StyledPrice priceDiff={priceDiff}>{priceDiff}</StyledPrice>
```

The way I have it now just felt the easiest to read for handling the three different cases and stylings.

src/components/SimplePriceChart/SimplePriceChart.tsx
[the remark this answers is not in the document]
kibagateaux: Ah interesting hack, thanks. The setTimeout also gets rid of the react warning so that should be a sufficient fix for now

kibagateaux:
Oh I totally got carried away and included INDEX page in here too.
I can create a new branch and rollback this one if you want. It's literally copy pasted from DPI code so you don't really have to review it.

src/components/ProductPage/ProductPageLayouts.tsx
```diff
@@ -0,0 +1,45 @@
...
+  :not(:last-child) {
+    border-bottom: 1px solid ${({ theme }) => theme.colors.grey[400]};
+  }
+`
+
+const SectionTitle = styled.h3`
+  font-size: 28px;
+  margin-top: 0;
+  color: ${({ theme }) => theme.textColor};
+`
```
controtie, line 45: interesting

src/views/DPI/components/DpiPriceChanges.tsx
controtie: What about this?

```
renderPriceChanges = () => {
  const priceDiff = calcDiff(days)

  if (!priceDiff) {
    return <>---<>
  }

  if (priceDiff > 0) {
    return <positive>....</positive>
  }

  return <negative>....</negative>
}
```
If you feel strongly about this, we can leave as is

src/components/ProductPage/ProductPriceChanges.tsx
```diff
@@ -0,0 +1,86 @@
...
+  flex-direction: column;
+`
+const StyledDateString = styled.div`
+  font-size: 16px;
+  margin-bottom: 16px;
+`
+
+const StyledNegativePrice = styled.div`
+  color: ${({ theme }) => {
+    console.log('theme', theme.colors.red, theme.colors)
```
controtie, line 77: lint

controtie (approved):
Looks very good. Let's merge once you've linted. I'll also put together some copy for the Index product page and we can get this out

src/views/DPI/components/DpiPriceChanges.tsx
kibagateaux: yea we can do that

PR #78, opened 2020-11-01
adds DPI index components from set api

kibagateaux:
[Warning]: Set Protocol API integration is actually untested, was getting CORS for localhost but i downloaded the data and mocked API response for development

Adds:
- `@types/lodash` package to use lodash in our code (lodash was already a dependency)
- `utils/fetchSetIndexComponents` API helper function
- `components/ProductPage/ProductIndexComponentsTable` for displaying index token data

Question - Justin said the API might change so I added this [error message](https://github.com/kibagateaux/index-ui/blob/b024337fa2f195a1830d6900e8df3fa90b84d617/src/utils/fetchSetIndexComponents.ts#L11). Any opinions on that?


kibagateaux:
Should be good to go @controtie

src/utils/fetchSetIndexComponents.ts
```diff
@@ -0,0 +1,32 @@
...
+    .then((response) => response.json())
+    .then((response) => {
+      if (!response?.portfolio?.components) {
+        // undocumented API endpoint. Throw error if not expected response format
+        throw new Error('Invalid API response from Set Protocol service')
+      }
+      const {
+        portfolio: { components },
+      } = response
+      const formattedComponents = components.map((component: any) => {
```
controtie, line 18: Can you add a note here on what this formatting is for?
Or break up the formatting logic into a named helper function

src/utils/fetchSetIndexComponents.ts
```diff
@@ -0,0 +1,32 @@
+import { camelCase } from 'lodash'
+
+const proxyPrefixURL = 'https://cors-anywhere.herokuapp.com/'
```
controtie, line 3: we'll probably end up removing this, but i'll handle it after testing the deploy

controtie (changes requested):
Looks mostly good to me. I'll try and get this merged today

src/utils/fetchSetIndexComponents.ts
kibagateaux: Yea sure. Right now it's only camelCasing strings but all the values are strings so wasn't sure if we want to format them to numbers here too. We use that `numeral` function on almost every input anyway so didn't seem necessary to.

src/utils/fetchSetIndexComponents.ts
kibagateaux: I tested fetch from the console on indexcoop.com and I also got a CORS from that domain so we still need it when deployed.
My understanding is that using this API endpoint is temporary until the subgraph is done and then this data will be available with minimal calculations on frontend right?

controtie: Ok we just ended up changing the set-core endpoint to allow CORS on all GET requests.

controtie:
Rebased and merged in #82

PR #83, opened 2020-11-11
Fix colors of percent changes in price graphs

ncitron:
The percent change metric is always green on the price graphs of DPI and INDEX. This PR changes that so that the percent change stat is red when at a loss, and green when at a gain.


controtie (approved):
LGTM

PR #88, opened 2020-11-14
fix: REP not displaying in DPI allocation list

ncitron:
## Changes
- Fix bug causing REP to not display in DPI token allocations
- Fix typo


controtie (approved):
LGTM

PR #93, opened 2020-11-24
Fix width of Integrations cards

ncitron:
## Changes
- Fixed width of integrations cards which were misaligned with the cards further up the page.
- Fixed typo

### DIFF
--- a/src/views/Home/components/Integrations.tsx
+++ b/src/views/Home/components/Integrations.tsx
@@ -17,7 +17,7 @@ const Integrations: React.FC = () => (
               <StyledCardIcon src='https://index-dao.s3.amazonaws.com/owl.png' />
               <StyledCardTitle>Index Coop</StyledCardTitle>
               <StyledCardDescription>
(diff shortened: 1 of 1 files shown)


controtie:
Do you have a screen shot of the alignment issues this PR addresses?

Also this PR changes the alignment of these bottom cards. Ideally we want 3 card per row
image

ncitron:
If you look closely, notice how the right edges of the rightmost cards are misaligned with the horizontal rule.

Screen Shot 2020-11-26 at 12 23 51 PM

I can fix the problem with the four cards in the last row. Would you prefer it look like this:

Screen Shot 2020-11-26 at 12 28 59 PM

Or this:

Screen Shot 2020-11-26 at 12 27 07 PM

controtie:
The first one looks good. Want to keep all the cards the same length. (imagine if there were two cards in the last row, might look weird if it was filled)

ncitron:
It should be good to go now.

controtie (approved):
LGTM

PR #98, opened 2020-12-03
Add buttons for 1D, 1W, and 1M price charts

ncitron:
## Changes
- Price charts have buttons for different timeframes
- On the 1D timeframe, dates are replaced with time using the `hh:mm` format

### DIFF
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -54,13 +54,15 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   const [price, setPrice] = useState(data)
 
   useEffect(() => {
-    if (durationSelector === 0) {
-      setPrice(data?.slice((data?.length * 29) / 30))
-    } else if (durationSelector === 1) {
-      setPrice(data?.slice((data?.length * 23) / 30))
-    } else if (durationSelector === 2) {
-      setPrice(data)
-    }
+    setTimeout(() => {
+      if (durationSelector === 0) {
+        setPrice(data?.slice((data?.length * 29) / 30))
+      } else if (durationSelector === 1) {
+        setPrice(data?.slice((data?.length * 23) / 30))
+      } else if (durationSelector === 2) {
+        setPrice(data)
+      }
+    }, 0)
   }, [durationSelector, data])
 
   const handleDailyButton = () => {
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -21,16 +21,17 @@ interface SimplePriceChartProps {
   }[]
   onMouseMove?: (...args: any[]) => any
   onMouseLeave?: (...args: any[]) => any
+  setIsDaily?: (...args: any[]) => any
 }
 
 const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   title,
   showTooltip,
-  showDurations,
   icon,
   data,
   onMouseMove = () => {},
   onMouseLeave = () => {},
+  setIsDaily = () => {},
 }) => {
   const theme = useTheme()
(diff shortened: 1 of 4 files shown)


src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -37,7 +40,42 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
...
-    return [new Date(x).toLocaleDateString(), '$' + formatFloats(y)]
+    let timeString = new Date(x).toLocaleDateString()
+    if (durationSelector === 0) {
+      timeString = new Date(x).toLocaleTimeString([], {
+        hour: 'numeric',
+        minute: 'numeric',
+      })
+    }
+    return [timeString, '$' + formatFloats(y)]
+  }
```
controtie, line 51: Maybe look into optimizing the above section? Could try memoizing `{ x: localeDateString }`

src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -37,7 +40,42 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
...
+
+  const [durationSelector, setDurationSelector] = useState<number>(2)
+  const [price, setPrice] = useState(data)
```
controtie, line 54: rename to something more descriptive like priceData

src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -104,4 +167,18 @@ const ChartTitle = styled.h2`
   font-size: 42px;
 `
 
+const DurationButton = styled.button`
+  background-color: radial-gradient(
+    circle at center top,
+    rgb(22, 20, 31),
+    rgb(22, 20, 31)
+  );
```
controtie, line 175: this radial gradient doesn't seem to be showing up very clearly? unless I'm missing something...
If not maybe remove?

src/views/DPI/components/MarketData.tsx
```diff
@@ -28,13 +30,35 @@ const MarketData: React.FC = () => {
...
+      if (isDaily) {
+        setDateString(
+          priceData.toLocaleTimeString([], {
+            hour: 'numeric',
+            minute: 'numeric',
+          })
+        )
+      } else {
+        setDateString(priceData.toDateString())
+      }
```
controtie, line 43: Also may look into optimizing the above calculation with memos

src/views/DPI/components/MarketData.tsx
```diff
@@ -28,13 +30,35 @@ const MarketData: React.FC = () => {
...
```
controtie, line 61: Ditto the above

controtie (changes requested):
Really good stuff! A couple notes:
1. Rebase off master. It looks like you're duplicating some commits on your card work from earlier which has been merged. Lmk if you need help with that
2. Take a look at the price charts on http://localhost:3000/dpi or http://localhost:3000/index
There's some expensive calculations in how you're handling price calculation which noticeably slows down the price chart component.

I've added some notes on places where you can look for speed improvements. But if you can fix the lagging price calculations in a different way thats fine.

Beyond that, looks pretty slick! 🦉

ncitron:
So I've done a bit of investigating on the lag issues for the charts. It appears to me that regardless of whether you run the current master or my PR, the graph appears to be laggy. This makes sense since both the current version and my version make repeated calls to `priceData.toDateString()`, which is what you proposed I optimize. This lag goes away however when you create an optimized production build using `yarn run build` and `serve -s build` to run it. This is why the version running in prod looks smooth. I am still happy to attempt to memoize it, but as far as I can tell this will cause no performance gains compared to the current version (as the current version does not memoize it). Let me know how you want me to proceed, and I'll push the other changes.

controtie:
Alright, I'll do a test deploy and get back to you

controtie:
Deploy looks good 👍

controtie (approved):

PR #107, opened 2021-01-05
Refactor rewards claim

ncitron:
This PR should make it much easier to add new months to the rewards. The relevant areas that need editing to add new months are in `src/contexts/Rewards/Provider.tsx` line 47 and `src/views/ContributorRewards/components/MonthsDropdown.tsx` line 72.


src/views/ContributorRewards/components/MonthlyClaim.tsx
```diff
@@ -16,12 +16,20 @@ const MonthlyClaim: React.FC = () => {
     return <Button disabled full text='Claim' variant='secondary' />
   }, [status, onClaimRewards, claimableQuantity])
 
+  const [currentMonth, setCurrentMonth] = useState('')
```
controtie, line 19: seems like you could replace this with by passing down `setMonth` and `rewardsMonth` from `useRewards()`

src/views/ContributorRewards/components/MonthlyClaim.tsx
```diff
@@ -16,12 +16,20 @@ const MonthlyClaim: React.FC = () => {
...
+    setMonth(currentMonth)
+  }, [setMonth, currentMonth])
+
   return (
     <>
       <Card>
         <CardContent>
           <Spacer size='sm' />
-          <MonthsDropdown />
+          <MonthsDropdown liftMonthState={liftMonthState} />
```
controtie, line 32: lift is kind of weird here. maybe `onSelectMonth`?

src/views/ContributorRewards/components/MonthsDropdown.tsx
```diff
@@ -1,22 +1,30 @@
...
   const theme = useTheme()
   const { pathname } = useLocation()
 
+  const [currentMonth, setCurrentMonth] = useState('December')
+  const setMonth = (label: string) => {
+    setCurrentMonth(label)
+  }
+  useEffect(() => {
+    props.liftMonthState(currentMonth)
+  }, [props, currentMonth])
```
controtie, line 17: same here, no need to duplicate setting state here if you can get it from the provider or via props. unless i'm missing something

src/contexts/Rewards/Provider.tsx
```diff
@@ -34,24 +42,42 @@ const Provider: React.FC = ({ children }) => {
     ethereum,
   }: { account: string | null | undefined; ethereum: provider } = useWallet()
 
+  //set proper contract address and merkle data depending on month
+  useEffect(() => {
+    switch (month) {
+      case 'December': {
+        break
+      }
+      case 'November': {
```
controtie, line 51: This thing will start breaking after 1 year because we'll have duplicate months. Maybe include the year

controtie (changes requested):
Looks mostly good, left some minor comments. I'll have the rewards contract + merkle soon, so get ready by removing the `test` naming. Ideally when this PR is ready it only requires updating the .env.prod and the december merkle file.

src/views/ContributorRewards/components/MonthlyClaim.tsx
```diff
@@ -16,12 +16,16 @@ const MonthlyClaim: React.FC = () => {
...
+    setMonth(month)
+  }
+
   return (
     <>
       <Card>
         <CardContent>
           <Spacer size='sm' />
-          <MonthsDropdown />
+          <MonthsDropdown onSelectMonth={onSelectMonth} />
```
controtie, line 28: Looks like `onSelectMonth` is unused since `MonthsDropdown` imports its own `useRewards`. Probably best to keep this here, add in month as well as a prop and remove `useRewards` from the `MonthsDropdown`

ncitron: Would it not be easier to just remove all references to `onSelectMonth`? I had originally been using the onSelectMonth strategy to lift the state of MonthsDropdown and it was much more code, which is why I changed it.

controtie: Sure, you could remove all references to `onSelectMonth` too. Just do one or the other cause atm it's not being used in `MonthsDropdown`. The case for keeping it as props is incase you want to reuse `MonthsDropdown` somewhere else but that case doesn't seem super likely

PR #113, opened 2021-01-12
transak integration

ncitron:
All that is needed for this to be merged into master is for `REACT_APP_TRANSAK_API_KEY` to be set to our proper API key and for the `environment` property on line 76 of `BuySellButton.tsx` to be set to `PRODUCTION`


controtie:
Merged in #130

PR #120, opened 2021-01-31
Fix google analytics on DPI/INDEX buy buttons

ncitron:
DPI and INDEX buy buttons now have differing classes. This should allow google tag manager to distinguish between the two.


controtie (approved):
LGTM

PR #141, opened 2021-03-03
fixed lint issues for lower CRA version and remove warning on browser…

ztcrypto:
… console too.


src/index.tsx
```diff
@@ -8,12 +8,7 @@ import gtmConfig from './utils/googleTagManager'
 
 TagManager.initialize(gtmConfig)
 
-ReactDOM.render(
-  <React.StrictMode>
```
controtie, line 12: what's the intended effect of removing strict mode?

src/views/CGI/CGI.tsx
```diff
@@ -143,15 +143,15 @@ const CgiProductPage = (props: { title: string }) => {
...
+            <div>
               The following rounding of data are used for the Index calculation
               and publication:
               <ul>
                 <li>Prices are rounded to eight significant figures</li>
                 <li>Weights are rounded to four decimal places</li>
                 <li>Index levels are rounded to two decimal places</li>
               </ul>
-            </p>
+            </div>
```
controtie, line 154: does this change how the page is rendered?

controtie: i.e. is the spacing still good/does it need to be fixed with further styling

controtie (commented):
looks mostly good. left some minor comments.
Also, why is yarn.lock updated but no packages in package.json have changed?

src/views/CGI/CGI.tsx
ztcrypto: it's good yeah

src/index.tsx
ztcrypto: just to remove browser console issue of unsafe_componentunmount warning.
actually it's ok to remain as it or remove.

controtie (approved):
LGTM

PR #148, opened 2021-03-14
Create FLI page with components and market data

evelyncc:
### Changes 
* Create FLI page with components and market data
* New method to fetch portfolio data to populate symbol, logo url, name, and index components

### To do
* Uniswap widget
* Add FLI to nav bar
* Update description
 
### Web
image

### Mobile
image


src/contexts/Balances/Provider.tsx
```diff
@@ -80,6 +83,9 @@ const Provider: React.FC = ({ children }) => {
       setDpiBalance(
         new BigNumber(balances[2]).dividedBy(new BigNumber(10).pow(18))
       )
+      setFliBalance(
+        new BigNumber(balances[2]).dividedBy(new BigNumber(10).pow(18))
```
controtie, line 87: need to increment the array index here and index for all balances afterwards (below)

.env.prod
```diff
@@ -3,7 +3,9 @@ REACT_APP_ENVIRONMENT=staging
 REACT_APP_ETHEREUM_RPC_URL=https://mainnet.eth.aragon.network/
 REACT_APP_SNAPSHOT_URL=https://hub.snapshot.page/api
 REACT_APP_TOKENSETS_API=https://api.tokensets.com/public
```
controtie, line 5: i guess we can deprecate this env in favor of `REACT_APP_TOKENSETS_URL`?

src/utils/tokensetsApi.ts
```diff
@@ -32,18 +32,31 @@ export const fetchSetComponents = (set: string) => {
...
-            ...comp,
-            [camelCase(k)]: component[k],
-          }),
-          {}
-        )
-        return camelCasedComponent
-      })
 
-      return formattedComponents
+      return formatComponents(components)
```
controtie, line 36: 👍

controtie (changes requested):
LGTM. Minor comments + 1 change on set balances.
And thanks for cleaning up some of our tech debt! 🥳

src/contexts/Balances/Provider.tsx
evelyncc: good catch!

.env.prod
evelyncc: that's the plan, unfortunately the `/public` makes it hard to reuse

controtie (approved):
Looks great, I'll handle the last bits like updating the set description once it's ready.

PR #157, opened 2021-04-01
TVL now includes all products, not just DPI

0xModene:
a new context was created to pool the TVLs together of all products, and used on the home page

### DIFF
--- a/src/constants/coingeckoIds.ts
+++ b/src/constants/coingeckoIds.ts
@@ -1,4 +1,4 @@
-export const coingeckoCgiId = 'cgi'
+export const coingeckoCgiId = 'coinshares-gold-and-cryptoassets-index-lite'
 export const coingeckoDpiId = 'defipulse-index'
-export const coingeckoFliId = 'ethfli'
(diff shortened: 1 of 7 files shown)


controtie (changes requested):
Hey, so it looks like the TVL displayed on this branch and the master branch isn't any different. To solve this is going to require a slightly larger refactor:

1. In general, we like to fetch our product's market data from coingecko
1. However, when CGI and FLI first launched they were brand new products and thus did not have market data available on coingecko. For the launch we fetched market historicals data from the tokensets api. You can find that in `FliTokenMarketDataProvider` or `CgiTokenMarketDataProvider`
1. To get the correct historicals + market cap data now, I suggest we move FLI and CGI market data fetching to use the coingecko api instead of tokensets.  
    Let's leave the tokensets fetching api in place, because we will likely go through the same process for future product launches.

0xModene:
There's a divergence in the TokenSets IDs and the CoinGecko IDs. Changes were made to account for that as well as what you mentioned @controtie

controtie (approved):
Looks great!

PR #161, opened 2021-04-06
Metaverse Index Support

controtie:
* Add product page for the Metaverse Index

### Desktop
image
image
image

### Mobile
image
image
image


controtie:
cc @dalberto

src/App.tsx
```diff
@@ -137,11 +143,15 @@ const Providers: React.FC = ({ children }) => {
                                     <DpiIndexComponentsProvider>
                                       <CgiTokenMarketDataProvider>
                                         <CgiIndexComponentsProvider>
-                                          <IndexTokenMarketDataProvider>
-                                            <SnapshotProposalsProvider>
-                                              {children}
-                                            </SnapshotProposalsProvider>
-                                          </IndexTokenMarketDataProvider>
+                                          <MviTokenMarketDataProvider>
```
dalberto, line 146: I'm curious, why is this wrapped inside the other components? Aren't these new components peers, of say `CgiTokenMarketDataProvider`? It may be a little confusing that the structure of the program doesn't match the hierarchy.

src/views/MVI/components/MarketData.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useState, useEffect } from 'react'
```
dalberto, line 1: Besides certain things like image assets, shouldn't this be pretty similar across different products? The reason I ask is because I think there may be an opportunity to abstract most of the functionality encapsulated by the component.

dalberto (commented):
This PR looks good overall, my comments and questions are more centered around maintainability and DRY-ness. I think there's an opportunity to build a data abstraction (probably an interface) that captures most of the functionality that a new product/token requires. I'm not sure what the timeline for this PR is, but I'd love to chat more about it, assuming you have the bandwidth.

src/App.tsx
controtie: Yup, all the market data/component providers are peers. Would love to dig in to how this might be better structured. It's not a problem right now but this is also leading to duplicate API calls since fetching happens in the provider

src/views/MVI/components/MarketData.tsx
controtie: Yup, definitely. I think we should be able to pass in price data, token image url to a single interface and get most of this stuff out

controtie:
@dalberto this PR is due to merge sometime today (MVI launch is tomorrow 12pm PST). If you're free this evening for a quick call let's go over it.

dalberto (approved):

PR #169, opened 2021-04-09
moved non-DPI indices off of the beta api

0xModene:
MVI, ETH2xFLI, and CGI were all using the beta tokensets API, which doesn't appear to have daily % change. For now they're using the release API


src/contexts/MediaQuery/MediaQueryProvider.tsx
```diff
@@ -15,8 +15,6 @@ const MediaQueryProvider: React.FC = ({ children }) => {
     setDeviceWidth(window.innerWidth)
     setDeviceHeight(window.innerHeight)
 
-    console.log('width is', window.innerWidth)
-
```
controtie, line 19: wow, thanks for removing this 🤦

controtie (approved):
awesome

PR #172, opened 2021-04-14
Fix Price Chart Data

0xModene:
- Data from CoinGecko now being pulled in 2 ways, once for the last 2 days in hourly interval, once in daily interval for max days available
- This should allow us to be more future ready for when a product goes beyond 30 days of age
- DPI data now actually getting full range of data, whereas before it was capped at 30 days, breaking the price charts
- These changes make it trivial to add more windows of time for the chart (3M, 1Y, etc) should we ever decide to do so

DPI 'Changes' section now has proper data to feed off of
image

### DIFF
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -19,6 +19,10 @@ interface SimplePriceChartProps {
     x: string | number
     y: number
   }[]
+  dayData?: {
+    x: string | number
+    y: number
+  }[]
   onMouseMove?: (...args: any[]) => any
   onMouseLeave?: (...args: any[]) => any
   setIsDaily?: (...args: any[]) => any
@@ -29,6 +33,7 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   showTooltip,
   icon,
   data,
+  dayData,
   onMouseMove = () => {},
   onMouseLeave = () => {},
   setIsDaily = () => {},
@@ -56,11 +61,11 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   useEffect(() => {
     setTimeout(() => {
       if (durationSelector === 0) {
-        setPrice(data?.slice((data?.length * 29) / 30))
+        setPrice(dayData?.slice(-24))
       } else if (durationSelector === 1) {
-        setPrice(data?.slice((data?.length * 23) / 30))
+        setPrice(data?.slice(-7))
       } else if (durationSelector === 2) {
-        setPrice(data)
+        setPrice(data?.slice(-30))
       }
     }, 0)
   }, [durationSelector, data])
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -19,7 +19,7 @@ interface SimplePriceChartProps {
     x: string | number
     y: number
   }[]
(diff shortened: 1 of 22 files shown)


src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -56,11 +61,11 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
   useEffect(() => {
     setTimeout(() => {
       if (durationSelector === 0) {
-        setPrice(data?.slice((data?.length * 29) / 30))
+        setPrice(dayData?.slice(-24))
```
controtie, line 64: why -24? Is this taking the last 24 days as default?

src/contexts/CgiTokenMarketData/CgiTokenMarketDataContext.ts
```diff
@@ -2,6 +2,7 @@ import { createContext } from 'react'
 
 interface CgiTokenMarketDataValues {
   prices?: number[][]
+  dayPrices?: number[][]
```
controtie, line 5: how do `dayPrices` differ from `prices`?
If we add support for 3M and 1Y prices in the future will there be `threeMonthPrices` and `oneYearPrices`?

controtie (commented):
Looks mostly OK. It seems like you've added extensibility to allow fetching price data from different date ranges in the future.

Which price data is used for which date ranges right now?
Testing the app it seems that we're displaying a lot less price data now. Some pruning of it makes sense, but I think we've pruned too much here.

controtie:
Current:
image

This PR:
image

src/components/SimplePriceChart/SimplePriceChart.tsx
0xModene: The first 2 days are pulled in hourly here, so -24 is indicating the last 24 hours. This can definitely be extended to include more days, and I'll match it to the chart that currently exists. Perhaps hourly for the first month, and beyond is daily? 

This sets hourly for most recent 48 hours, then daily beyond that.

src/contexts/CgiTokenMarketData/CgiTokenMarketDataContext.ts
0xModene: prices is everything that CG will give us, in daily intervals. dayPrices is hourly for the most recent 48 hours. 

I'll gladly take better name suggestions, especially if the dayPrices is turning into a week or more to match current granularity. Will await your suggestions

src/components/SimplePriceChart/SimplePriceChart.tsx
controtie: Got it makes sense. Lets see what hourly for 48 hours + daily beyond that looks like. It sounds like it should be fine.

src/contexts/CgiTokenMarketData/CgiTokenMarketDataContext.ts
controtie: I see. yeah maybe `oneWeekPrices` or `oneMonthPrices` depending on which you go with.

0xModene:
How's this?

image

controtie (approved):
Looks great 👍

controtie:
actually looks like there might be a regression on the home page. gonna take a look at it later today

0xModene:
Ah crap, that's an oversight on my end. Sending a patch now, I know exactly what this is

PR #179, opened 2021-04-17
fixed price change epoch bug

0xModene:
Price Change epoch now set for the data range selected on the chart, where previously it was change since start

### DIFF
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -25,7 +25,7 @@ interface SimplePriceChartProps {
   }[]
   onMouseMove?: (...args: any[]) => any
   onMouseLeave?: (...args: any[]) => any
-  setIsDaily?: (...args: any[]) => any
+  setChartRange?: (...args: any[]) => any
 }
 
 const MarketDataChart: React.FC<SimplePriceChartProps> = ({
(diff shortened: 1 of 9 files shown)


0xModene:
These changes _should_ be identical for every MarketData except the homepage one, which is minimally different, just doesn't need to account for changing price epoch

src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -69,19 +69,19 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
...
   }
   const handleWeeklyButton = () => {
     setDurationSelector(1)
-    setIsDaily(false)
+    setChartRange(7)
   }
   const handleMonthlyButton = () => {
     setDurationSelector(2)
-    setIsDaily(false)
+    setChartRange(30)
```
controtie, line 84: small knit: using an enum for these values might improve readibility

```
enum PriceChartRangeOption {
  DAILY_PRICE_RANGE = 1,
  WEEKLY_PRICE_RANGE = 7,
  MONTHLY_PRICE_RANGE = 30,
}

setChartRange(PriceChartRangeOption.DAILY_PRICE_RANGE)
```

controtie (approved):
Awesome work 🎉

src/components/SimplePriceChart/SimplePriceChart.tsx
0xModene: That's a great point, easy change to make as well.

PR #180, opened 2021-04-20
Readability enums

0xModene:
Changed durationSelector and chartRange values to enums for readibility purposes as mentioned [here](https://github.com/SetProtocol/index-ui/pull/179#discussion_r616887004)

### DIFF
--- a/src/components/SimplePriceChart/SimplePriceChart.tsx
+++ b/src/components/SimplePriceChart/SimplePriceChart.tsx
@@ -46,7 +46,7 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
       payload: { x, y },
     } = chartData
     let timeString = new Date(x).toLocaleDateString()
-    if (durationSelector === 0) {
+    if (durationSelector === Durations.DAILY) {
       timeString = new Date(x).toLocaleTimeString([], {
         hour: 'numeric',
         minute: 'numeric',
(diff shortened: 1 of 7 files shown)


src/components/SimplePriceChart/SimplePriceChart.tsx
```diff
@@ -55,33 +55,59 @@ const MarketDataChart: React.FC<SimplePriceChartProps> = ({
...
+    DAILY = 0,
+    WEEKLY = 1,
+    MONTHLY = 2,
+  }
+
+  enum PriceChartRangeOption {
+    DAILY_PRICE_RANGE = 1,
+    WEEKLY_PRICE_RANGE = 7,
+    MONTHLY_PRICE_RANGE = 30,
+  }
```
controtie, line 68: It seems like we'd want to use these enums in the `MarketData.tsx` files as well.
e.g. https://github.com/SetProtocol/index-ui/pull/179/files#diff-9c164f14735179677e274d6c9b6c118c8b939c9a572ee3dbda9492ee53534b21R13
or 
https://github.com/SetProtocol/index-ui/pull/179/files#diff-9c164f14735179677e274d6c9b6c118c8b939c9a572ee3dbda9492ee53534b21R37

We don't have a great structure for enums right now, but you could create a `priceChart` in the `/constants` folder to import to all related files

controtie (changes requested):
good stuff, I think we want to use those enums in a couple more places

src/components/SimplePriceChart/SimplePriceChart.tsx
0xModene: I think I've got this covered now in [the latest commit](https://github.com/SetProtocol/index-ui/pull/180/commits/b06cbeb320b30c8591f349770da608955a38aef1)

controtie (approved):

PR #181, opened 2021-04-21
Market data update

0xModene:
Added Market Data for all products to the homepage
image
image

### DIFF
--- a/src/views/Home/components/MarketData.tsx
+++ b/src/views/Home/components/MarketData.tsx
@@ -10,17 +10,11 @@ import SimplePriceChart from 'components/SimplePriceChart'
 
 import useDpiTokenMarketData from 'hooks/useDpiTokenMarketData'
 import { productTokensBySymbol } from 'constants/productTokens'
+import useAllProductsMarketData from 'hooks/useAllProductsMarketData'
+import { PriceChartRangeOption } from 'constants/priceChartEnums'
 
 const MarketData: React.FC = () => {
-  const {
-    latestVolume,
-    latestMarketCap,
-    latestPrice,
-    prices,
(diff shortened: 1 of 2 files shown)


src/views/Home/components/MarketData.tsx
```diff
@@ -1,72 +1,176 @@
...
-  } = useDpiTokenMarketData()
-  const priceAtEpochStart = prices?.slice(-30)[0]?.[1] || 1
-  const epochPriceChange = (latestPrice || 0) - priceAtEpochStart
-  const dpiTokenIcon = {
-    src: 'https://index-dao.s3.amazonaws.com/defi_pulse_index_set.svg',
-    alt: 'DefiPulse Index Logo',
+  const allMarketData = useAllProductsMarketData()
+
+  const [indexSelector, setIndexSelector] = useState<number>(0)
+  const [indexMetaData, setIndeexMetaData] = useState<{
```
controtie, line 19: index can be confusing since it also can refer to an array index... I've been guilty of it in the past but we should make an effort to move towards less ambiguous language.

Something like `productMetaData`/`setProductMetaData` seems good here.

controtie (changes requested):
Looks mostly good! Just a small typo + some minor naming change requested.

src/views/Home/components/MarketData.tsx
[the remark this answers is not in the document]
0xModene: 🤦‍♂️

0xModene: Great to know for the Big Refactor story that's upcoming

controtie (approved):

PR #182, opened 2021-04-23
[Bugfix] Clicking on backdrop was not closing the WalletModal

pouta:
The onDismiss function was not being passed as prop to the CustomModal component.


controtie (approved):
LGTM!

PR #185, opened 2021-04-28
staking apy update

0xModene:
staking apy math is a bit more accurate now, basing the total off of what's being staked, not whats in Uniswap

### DIFF
--- a/src/contexts/Prices/PricesProvider.tsx
+++ b/src/contexts/Prices/PricesProvider.tsx
@@ -5,13 +5,25 @@ import { useQuery } from '@apollo/react-hooks'
 import PricesContext from './PricesContext'
 
 import { DPI_ETH_UNISWAP_QUERY, ETH_MVI_UNISWAP_QUERY } from 'utils/graphql'
-import { indexTokenAddress } from 'constants/ethContractAddresses'
+import {
+  farmTwoAddress,
+  indexTokenAddress,
+  mviStakingRewardsAddress,
+} from 'constants/ethContractAddresses'
+import { getAmountOfStakedTokens } from 'index-sdk/stake'
+import useWallet from 'hooks/useWallet'
 
 const PricesProvider: React.FC = ({ children }) => {
   const [indexPrice, setIndexPrice] = useState<string>('0')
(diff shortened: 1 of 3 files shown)


src/contexts/Prices/PricesProvider.tsx
```diff
@@ -67,32 +83,74 @@ const PricesProvider: React.FC = ({ children }) => {
...
 
-    setMviRewardsApy(calculatedApy.toFixed(2))
-  }, [usdInEthMviPool, indexPrice])
+    //get usd price per lp token
+    const pricePerLPToken = new BigNumber(usdInEthMviPool).dividedBy(
+      new BigNumber(totalSupplyInEthMviPool)
+    )
+
+    //multiply by totalSupply
+    getAmountOfStakedTokens(ethereum, mviStakingRewardsAddress).then(
```
controtie, line 140: Are there any problems if `getAmountOfStakedTokens` request fails? Do we need to catch anything?

controtie (approved):
Looks mostly good. Something to think on: how can we improve separation of concerns in these providers. Not sure if that's against the ethos of react hooks but some of these files can get busy.

src/contexts/Prices/PricesProvider.tsx
0xModene: Great catch, I'll provide some error handling today. There's not a lot we can do data-wise if that happens, but we can zero out everything and log an error.

src/contexts/Prices/PricesProvider.tsx
```diff
@@ -113,8 +113,11 @@ const PricesProvider: React.FC = ({ children }) => {
         const calculatedApy = dailyYield.multipliedBy(365)
 
         setFarmTwoApy(calculatedApy.toFixed(2))
-      }
-    )
+      })
+      .catch((error) => {
+        console.log(error)
+        setFarmTwoApy('0.00')
```
0xModene, line 119: @controtie How's this? Not much else we can do here except fail gracefully

controtie: Yeah I think that should be good for now

PR #199, opened 2021-05-02
created generic MarketData component

### LINKED ISSUE #198: Consolidate MarketData components
We only need one MarketData component for the product pages, where currently we use a new one for each page. Make the code reusable.

0xModene:
This is intended to resolve #198 by consolidating MarketData into a single component


0xModene:
This is not meant to be used on the homepage as this component is a little different than the others. There will likely be another change later to break these different bits out and truly unify MarketData, but I want to keep this PR as small as possible

controtie (approved):
LGTM

PR #201, opened 2021-05-02
token metadata is all in one place now

### LINKED ISSUE #200: Token data structure needs consolidation
The token data is kind of everywhere, and this increases pain when adding new products to the site. The token metadata should be in one place, so as to make it easier to discover and prevent data mismatch issues.

0xModene:
This should resolve #200 , better centralizing token metadata


controtie (approved):
LGTM!

PR #209, opened 2021-05-06
add /join and /discord as redirect to discord

jierlich:
Add redirect links to discord for ease of access.

As a heads up, I can't run locally without updating `caniuse-lite`. Should I include the update for this here as well?

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,6 +50,8 @@ import createTheme from 'utils/createCustomTheme'
 import graphqlClient from 'utils/graphql'
 import { RewardsProvider } from './contexts/Rewards'
 
+const DISCORD_LINK = 'https://discord.com/invite/74xXr57XHT'
+
 const App: React.FC = () => {
   const [mobileMenu, setMobileMenu] = useState(false)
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -50,6 +50,8 @@ import createTheme from 'utils/createCustomTheme'
 import graphqlClient from 'utils/graphql'
 import { RewardsProvider } from './contexts/Rewards'
 
+const DISCORD_LINK = 'https://discord.com/invite/74xXr57XHT'
```
dalberto, line 53: @controtie We should consider where we want to put constants like this. `App.tsx` seems like a suboptimal place.

jierlich: Good point. It looks like `src/constants` already exists. What are your thoughts on a `src/constants/externalLinks.ts` file?

dalberto: That seems like a good place, but I defer to @controtie.

controtie: src/constants/externalLinks sounds great!

jierlich: Cool! Just pushed a change adding that file + a small stylistic cleanup

0xModene (approved):
LGTM

I went ahead and reviewed this as this was mentioned on the DeFridays call and wanted to make sure it was out there

PR #212, opened 2021-05-07
[WIP] Sushiswap Trade Support

controtie:


src/uniswap-sdk/uniswap.ts
```diff
@@ -29,19 +32,31 @@ export const getUniswapRouterContract = (provider: provider) => {
...
+const getSushiswapRouterContract = (provider: provider) => {
+  const web3 = new Web3(provider)
+  const contract = new web3.eth.Contract(
+    (UniswapRouterABI.abi as unknown) as AbiItem,
+    sushiswapRouterAddress
+  )
+  return contract
+}
+
 export const getUniswapTradeTransaction = (
```
0xModene, line 44: Note: maybe we generalize this a bit since it looks like we're going to be working a bit more with Sushi as time goes on?

Maybe that's worth doing in another PR, and can include the filename

0xModene (approved):
LGTM

In a future PR lets rename some things to be more generic, since we're including Sushi now. That could be a good janitorial good-first-issue task

src/uniswap-sdk/uniswap.ts
controtie: yup, sounds good

PR #219, opened 2021-05-09
Modular product UI

0xModene:
The ProductPage UI is now modular

- All Product Pages use the same UI component
- Product Page Components (ie `DPI.tsx`) are now essentially just landing pages for data combination
- Description copy still needs to be created, but needs to be added as a child to the ProductDataUI component
- Included BTC FLI in this change, but still disabled from view
- Created one IndexComponent interface so all context providers use the same interface, instead of replicating it through all providers
- no more hard coded token IDs throughout wallet/metamask features, instead using ProductToken objects


controtie (changes requested):
small nits, but overall very nicely done!
can you rebase this? I'll get it merged in the morning.
I just did one and it seemed straight forward, but i can't figure out how to push to your branch 😅 

Also: something we can start thinking about is what simple testing frameworks can we add while the codebase is relatively simple. I tested this PR but it's possible I missed something. I'll try and put some thoughts down later this week.

controtie (changes requested):
one more thing: the INDEX token currently doesn't have a `tokensetsId` which makes sense since it's not really a token sets product. But we _do_ use it to fetch trade quotes + the symbol is used in the buy/sell widget, so might as well add it in.

image

0xModene:
> one more thing: the INDEX token currently doesn't have a `tokensetsId` which makes sense since it's not really a token sets product. But we _do_ use it to fetch trade quotes + the symbol is used in the buy/sell widget, so might as well add it in.
> 
> image

this has been updated, and the INDEX change shows the response to both
image
image

controtie (approved):
LGTM

PR #220, opened 2021-05-10
NAV update

### LINKED ISSUE #188: Display product NAV
~~we need to display premium/discount calculation on the buy/sell process between NAV and DEX price~~

This will actually be changed to displaying the NAV on the product page

0xModene:
Unified Product UI now shows NAV in Stats section. This should resolve #188


src/components/ProductPage/ProductTokenStats.tsx
```diff
@@ -11,13 +12,15 @@ interface ProductTokenStatsProps {
   fees?: {
     streamingFee: string
   }
+  nav: number
```
controtie, line 15: I think (in general) we should use longer, more descriptive variable names over short ones where possible. i think a lot of people might confuse this as having to do with navigation. just `netAssetValue` would be helpful in reading the code.

src/components/ProductPage/ProductDataUI.tsx
```diff
@@ -50,6 +54,17 @@ const ProductDataUI: React.FC<ProductDataUIProps> = ({
     if (value) setReferral(value)
   }, [value, setReferral])
 
+  const navReducer = (nav: number, component: IndexComponent): number => {
+    return nav + parseFloat(component.totalPriceUsd)
+  }
+
+  const getNav = () => {
+    let acc = 0
```
controtie, line 62: `acc` doesn't feel super descriptive/useful here. I would prefer to either remove it (put the 0 inline) or rename it

src/components/ProductPage/ProductDataUI.tsx
```diff
@@ -50,6 +54,17 @@ const ProductDataUI: React.FC<ProductDataUIProps> = ({
     if (value) setReferral(value)
   }, [value, setReferral])
 
+  const navReducer = (nav: number, component: IndexComponent): number => {
+    return nav + parseFloat(component.totalPriceUsd)
```
controtie, line 58: if totalPriceUsd is undefined will this break?

src/components/ProductPage/ProductTokenStats.tsx
```diff
@@ -51,6 +54,10 @@ const ProductTokenStats: React.FC<ProductTokenStatsProps> = ({
           </StyledStatMetric>
         </StyledStat>
         {streamingFee}
+        <StyledStat>
+          <StyledStatTitle>NAV</StyledStatTitle>
```
controtie, line 58: Let's label this as Net Asset Value as well

controtie (changes requested):
Minor changes to code style + labeling.

Also the calculated net asset value from tokensets looks like it diverges from coingecko quite a bit. Not sure how coingecko is e.g. calculating eth/fli or dpi prices. Let's see what the response from users looks like.

src/components/ProductPage/ProductTokenStats.tsx
0xModene: I'll change this everywhere

controtie (approved):
LGTM!

PR #239, opened 2021-05-19
dependency and build warning cleanup

### LINKED ISSUE #183: High risk vulnerabilities in dependencies
As the website integrates with the user wallet, we must be very careful to not introduce any attack vectors that could put our users funds in danger.

After running npm audit I found one that should be solved ASAP:

web3 package has a vulnerability that may leak the private keys. The [vulnerability page](https://npmjs.com/advisories/877) states the following:    
> Exploiting this vulnerability likely requires a Cross-Site Scripting vulnerability to access the private key.

Other notable vulnerabilities:

- axios - [Server Side Request Forgery](https://npmjs.com/advisories/1594)
- elliptic - [Usage of broken or risky crypto algos](https://npmjs.com/advisories/1648) and [Signature Malleability]( https://npmjs.com/advisories/1547)

0xModene:
This should resolve #183 along with some other issues


src/constants/productTokens.ts
```diff
@@ -89,11 +89,13 @@ export const productTokensBySymbol = {
...
+const ProductTokens = [
   DefiPulseIndex,
   MetaverseIndex,
   CoinsharesCryptoGoldIndex,
   Ethereum2xFlexibleLeverageIndex,
   IndexToken,
   Bitcoin2xFlexibleLeverageIndex,
 ]
+
+export default ProductTokens
```
0xModene, line 101: It came as a warning on build to put default exports into a variable and then export

src/contexts/BuySell/BuySellProvider.tsx
```diff
@@ -58,26 +61,40 @@ const BuySellProvider: React.FC = ({ children }) => {
...
+    dpiBalance,
+    ethfliBalance,
+    btcfliBalance,
+    cgiBalance,
+    mviBalance,
+    ethBalance,
+    daiBalance,
+    usdcBalance,
+    selectedCurrency?.id,
+  ])
```
0xModene, line 97: It's standard practice to put something like this into a memo since the IO is fixed, in a way, and post-dependabot this was being flagged in a warning

tsconfig.json
```diff
@@ -14,7 +18,10 @@
...
     "noEmit": true,
     "jsx": "react",
-    "baseUrl": "./src"
+    "baseUrl": "./src",
+    "noFallthroughCasesInSwitch": true
   },
-  "include": ["src"]
+  "include": [
+    "src"
+  ]
```
0xModene, line 26: this is all the result of an npm audit

PR #300, opened 2021-07-01
Use 0x API for trades

ncitron:
# **Pull Request Template**

## **Type of Change**

*Select a category that relates to the content of your pull request (choose one).*

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [x] Code Refactoring
- [ ] Update of code styling

&nbsp;

## **Does this request include a breaking change?**

*Is there any potential for this change to crash Index UI or make it unusable?*

- [ ] Yes
- [x] No

&nbsp;

## **Description**

Link to specification: https://docs.google.com/document/d/1V9Fmdn0JQwEyvqWB5toi-nhH8PgE9fwun18fezgaOeU/edit?usp=sharing 

&nbsp;

**Summary of Changes:**
- use 0x's API to route trades
- remove premium in order summary (it was misleading, as it was the slippage, not the premium)

## **Checklist**

*Please review and confirm each of the following before creating your pull request.*

- [x] I have reviewed the [Contribution Guidelines](https://github.com/SetProtocol/index-ui/blob/master/CONTRIBUTING.md).
- [x] I have performed a self-review of my own code.
- [x] I have checked my code and corrected any errors.
- [x] I have commented my code or used an effective naming scheme.
- [x] I have updated my repository to match the master at `SetProtocol/index-ui`.
- [x] I have included test data or screenshots that prove my fix is effective or that my feature works.

&nbsp;

*If all of the above are **true**, then you may mark your pull request as **ready for review**.*


src/components/BuySell/components/BuySellButton.tsx
```diff
@@ -4,16 +4,9 @@ import { RoundedButton } from 'components/RoundedButton'
...
-  daiTokenAddress,
-  usdcTokenAddress,
-  dpiTokenAddress,
-  eth2xfliTokenAddress,
-  btc2xfliTokenAddress,
-  indexTokenAddress,
-  uniswapRouterAddress,
-  sushiswapRouterAddress,
+  ethTokenAddress,
+  zeroExRouterAddress,
```
0xModene, line 9: So everything goes through 0x now? Such tidy, much wow

ncitron: Yep! One step closer to migrating to V3...

controtie (changes requested):
Can you leave some notes about the 0x API in this PR? Like what the payload pieces are, what it means for us to submit it directly with web3.transaction etc... It's hard to dig in and tell if the implementation is correct. We should have specced this out a bit more before starting work.

I think there's also some minor UI bugs like all currency options being highlighted:
image

And the please confirm your transaction message still being shown after the transaction's been submitted:
image

src/components/BuySell/components/BuySellButton.tsx
```diff
@@ -147,10 +53,28 @@ const BuySellButton: React.FC = () => {
     buttonAction = onExecuteBuySell
   }
 
+  if (
+    zeroExTradeData?.sellTokenAddress === ethTokenAddress ||
+    zeroExTradeData?.sellTokenAddress === undefined
```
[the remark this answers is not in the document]
ncitron: Yeah the undefined case is to prevent it from displaying an approve tokens message when the user is not logged in

src/components/BuySell/components/MaxButton.tsx
```diff
@@ -10,56 +10,50 @@ const MaxButton: React.FC = () => {
...
     }
-    requiredQuantity = new BigNumber(uniswapData?.amount_in || 0).dividedBy(
-      new BigNumber(10).pow(6)
-    )
+    requiredQuantity = new BigNumber(zeroExTradeData?.maxInput || 0)
   }
 
   const userHasSufficientFunds = spendingTokenBalance.isGreaterThanOrEqualTo(
-    requiredQuantity
+    requiredQuantity.multipliedBy(0.99999) // accounts for rounding errors
```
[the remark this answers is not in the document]
ncitron: The root of this problem is that the useBalances hook returns balances in displayable values, rather than their precise (1e18) values. We end up losing some precision when balances get rounded during this conversion. We should definitely, think about refactoring the useBalances hook to fix this in the future, but it is quite deeply integrated into other parts of the codebase, so it probably deserves a separate PR

src/components/BuySell/components/BuySellButton.tsx
```diff
@@ -147,10 +56,15 @@ const BuySellButton: React.FC = () => {
     buttonAction = onExecuteBuySell
   }
 
+  if (loginRequired) {
+    buttonText = 'Login'
+    buttonAction = onOpenWalletModal
+  }
+
```
controtie, line 63: very nice

src/components/BuySell/components/MaxButton.tsx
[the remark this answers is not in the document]
controtie: Makes sense, can you write something to that effect in the comment so we have that context in the future? maybe mention we lose precision in `BalancesProvider` and this rounds down to account for that.

src/contexts/BuySell/BuySellProvider.tsx
```diff
@@ -1,38 +1,28 @@
...
 import useBalances from 'hooks/useBalances'
 import useTransactionWatcher from 'hooks/useTransactionWatcher'
-import { useDebounce } from 'hooks/useDebounce'
-import { getUniswapTradeTransaction } from 'uniswap-sdk/uniswap'
-import {
-  getUniswapTradeType,
-  getUniswapCallData,
-  getUniswapTransactionOptions,
-} from './utils'
+import { getZeroExTradeData } from '../../utils/zeroExUtils'
```
controtie, line 10: just `utils/zeroExUtils` should work here, similar to the bignumber import above

src/utils/coingeckoApi.ts
```diff
@@ -27,3 +29,28 @@ export const fetchHistoricalTokenMarketData = (
...
+
+  const getPriceUrl =
+    baseURL +
+    `/simple/token_price/ethereum/?contract_addresses=${address}&vs_currencies=${baseCurrency}`
+
+  const resp = await fetch(getPriceUrl)
+  const data = await resp.json()
+  if (!data[address.toLowerCase()]) return 0
+  return data[address.toLowerCase()][baseCurrency]
+}
```
controtie, line 56: remove console.logs. Also do we need try/catch blocks in here? if the coingecko request fails how does that affect the rest of the flow?

src/utils/coingeckoApi.ts
```diff
@@ -27,3 +29,28 @@ export const fetchHistoricalTokenMarketData = (
     })
     .catch((error) => console.log(error))
 }
+
+export const fetchCurrentPrice = async (
```
controtie, line 33: Rename to `fetchCoingeckoTokenPrice`

src/utils/zeroExUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+    params.buyAmount = getDecimalAdjustedAmount(
+      buySellAmount,
+      tokenInfo[buyToken].decimals
+    )
+  }
+
+  return params
+}
+
+const processApiResult = async (
```
controtie, line 72: can we add a comment in here detailing what this does? is it mostly formatting for people to read? does this affect the payload that we submit to 0x?

src/utils/zeroExUtils.ts
```diff
@@ -0,0 +1,143 @@
...
+  zeroExData.minOutput = isExactInput
+    ? guaranteedPrice * zeroExData.displaySellAmount
+    : parseFloat(buySellAmount)
+  zeroExData.maxInput = isExactInput
+    ? parseFloat(buySellAmount)
+    : guaranteedPrice * zeroExData.displayBuyAmount
+
+  zeroExData.formattedSources = formatSources(zeroExData.sources)
+
+  const buyTokenPrice = await fetchCurrentPrice(tokenInfo[buyToken].address)
```
controtie, line 98: Add a TODO, we should decouple this in a future PR

controtie (changes requested):
Nice work on the previous changes, I've left some more comments. Also check out your token pricing stuff, something seems to be off when selling:

image

src/contexts/BuySell/BuySellProvider.tsx
```diff
@@ -1,38 +1,29 @@
...
-} from './utils'
+import { getZeroExTradeData } from 'utils/zeroExUtils'
 import trackReferral from 'utils/referralApi'
 import { waitTransaction } from 'utils/index'
 import { TransactionStatusType } from 'contexts/TransactionWatcher'
-import { Bitcoin2xFlexibleLeverageIndex } from 'constants/productTokens'
 import { currencyTokens } from 'constants/currencyTokens'
-import { UniswapPriceData } from './types'
+import { ZeroExData } from './types'
+import { reject } from 'lodash'
```
controtie, line 16: I think this is a mistaken import?

src/contexts/BuySell/BuySellProvider.tsx
```diff
@@ -68,105 +59,73 @@ const BuySellProvider: React.FC = ({ children }) => {
...
 
     try {
       onSetTransactionStatus(TransactionStatusType.IS_APPROVING)
 
-      const transactionId = await uniswapTradeTransaction()
+      const transactionId: string = await new Promise((resolve, reject) => {
+        tx.on('transactionHash', (txId: string) => {
+          if (!txId) reject()
+          resolve(txId)
+        }).once('error', () => {
```
controtie, line 123: What's this do?

src/utils/zeroExUtils.ts
```diff
@@ -0,0 +1,144 @@
...
+    params.buyAmount = getDecimalAdjustedAmount(
+      buySellAmount,
+      tokenInfo[buyToken].decimals
+    )
+  }
+
+  return params
+}
+
+// Adds some additional information to the ZeroExData return object. This extra information is only used for display purposed, and
```
controtie, line 68: display purposes*

.env.prod
```diff
@@ -11,6 +11,7 @@ REACT_APP_INDEX_TOKEN_ADDRESS=0x0954906da0Bf32d5479e25f46056d22f08464cab
 REACT_APP_DAI_TOKEN_ADDRESS=0x6B175474E89094C44Da98b954EedeAC495271d0F
 REACT_APP_USDC_TOKEN_ADDRESS=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
 REACT_APP_BTC2X_FLI_TOKEN_ADDRESS=0x0B498ff89709d3838a063f1dFA463091F9801c2b
+REACT_APP_ETH_TOKEN_ADDRESS=0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
```
controtie, line 14: Is this canonical? Or is this specific to 0x's API?

controtie: It seems like this address is only used in relation to 0x API. If it's not canonical we should label it as such. Maybe include it as a constant in `ethContractAddresses` instead of setting it as ENV

controtie (changes requested):
Just a couple more minor fixes. Will do a thorough test once these are in

src/contexts/BuySell/BuySellProvider.tsx
ncitron: This is part of the logic for getting to the transaction failed screen. It justs sets the promise result to "0" if a tx fails, which is then checked for and thrown when it does fail (being caught at the bottom of the try/catch that this is in

controtie: In this case can we just `reject()` like on 121? would also remove the need for line 128

src/utils/coingeckoApi.ts
```diff
@@ -27,3 +29,37 @@ export const fetchHistoricalTokenMarketData = (
...
+): Promise<number> => {
+  if (address === ethTokenAddress) {
+    const getPriceUrl =
+      baseURL + `/simple/price/?ids=ethereum&vs_currencies=${baseCurrency}`
+
+    const resp = await fetch(getPriceUrl)
+
+    const data = await resp.json().catch((err) => {
+      return 0
+    })
```
controtie, line 45: This pattern looks a bit weird, but I tested it and it seems OK.

src/contexts/Balances/Provider.tsx
```diff
@@ -108,52 +108,64 @@ const Provider: React.FC = ({ children }) => {
       ])
 
       setEthBalance(
-        new BigNumber(balances[0]).dividedBy(new BigNumber(10).pow(18))
+        new BigNumber(balances[0]).minus(1).dividedBy(new BigNumber(10).pow(18))
```
controtie, line 111: Let's hold off on this fix actually.

controtie (changes requested):
Last couple things.

controtie (approved):
LGTM!

src/utils/zeroExUtils.ts
```diff
@@ -0,0 +1,148 @@
...
+  buyToken: string,
+  buySellAmount: string
+): any => {
+  const params: any = {
+    sellToken: tokenInfo[sellToken].address,
+    buyToken: tokenInfo[buyToken].address,
+  }
+
+  if (isExactInput) {
+    params.sellAmount = getDecimalAdjustedAmount(
```
justinkchen, line 54: @ncitron can you explain to me why isExactInput is what differentiates if we use sellAmount or buyAmount?

PR #304, opened 2021-07-05
Add 2x total supply hooks

Boda805:
# **Pull Request Template**

## **Type of Change**

*Select a category that relates to the content of your pull request (choose one).*

New feature

## **Does this request include a breaking change?**

*Is there any potential for this change to crash Index UI or make it unusable?*

Unlikely

## **Description**

Purpose is to not rely on env variables for 2x total supply, these hooks grab the data directly from SupplyCapIssuance contracts

*If you are resolving an issue, please link the corresponding issue number.*
No known issue number, work assigned through dylan on dev chat in Discord

*If all of the above are **true**, then you may mark your pull request as **ready for review**.*


Boda805:
image

0xModene:
This looks mostly good, but a small nit

Can you remove these lines from `.env.prod`?


```
REACT_APP_ETH2X_FLI_SUPPLY_CAP=600000
REACT_APP_BTC2X_FLI_SUPPLY_CAP=400000
```

0xModene:
Also in `components/ProductPage/ProductTokenStats.tsx` updated `formattedSupplyCap` to be something like this: 

```
  const formattedSupplyCap = supplyCap
    ? numeral(supplyCap?.toString() || '0').format('0,0')
    : '--'
```

This way it doesn't show incorrect data until a user connects their wallet

0xModene (changes requested):
looks mostly good, 2 small updates are needed though

Boda805:
Thanks for reviewing, I realized I forgot to remove env variables right after I did PR, but anyway this should do the trick.

src/contexts/Btc2xFliTokenSupplyCap/Btc2xFliTokenSupplyCapProvider.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useState, useEffect } from 'react'
+import TotalSupplyContext from './Btc2xFliTokenSupplyCapContext'
+import { getSupplyCap } from 'utils'
+import { btc2xfliSuppyCapAddress } from 'constants/ethContractAddresses'
+import useWallet from 'hooks/useWallet'
+import SupplyCapIssuanceABI from 'index-sdk/abi/SupplyCapIssuanceHook.json'
+import { AbiItem } from 'web3-utils'
+import { provider } from 'web3-core'
```
controtie, line 8: small nit: group imports for external packages together at the top

controtie: Also group utils together etc...

src/utils/index.ts
```diff
@@ -139,3 +139,21 @@ export const getTotalSupply = async (
     return '0'
   }
 }
+
+export const getSupplyCap = async (
+  tokenAddress: string,
+  abi: AbiItem,
```
controtie, line 145: I don't think you need to pass in the ABI here. Just import the ABI at the top of the file since the abi required to `getSupplyCap` shouldn't change.

controtie: So then this should only accept token address + provider

src/views/BTC2XFLI/BTC2XFLI.tsx
```diff
@@ -30,13 +33,13 @@ const Btc2xFliProductPage = (props: { title: string }) => {
...
     balance: btcfliBalance,
-    supplyCap: supplyCap,
+    supplyCap: btcfliSupplyCap,
     currentSupply: btcfliTotalSupply,
   }
   const { account } = useWallet()
 
   const isApproachingSupplyCap = btcfliTotalSupply
-    ?.div(supplyCap)
+    ?.div(btcfliSupplyCap as BigNumber)
```
controtie, line 42: is there a scenario here where we're dividing by 0? and would that break anything?

controtie (changes requested):
Awesome work! This is it's like 95% there. Just left a couple code styling considerations

src/views/BTC2XFLI/BTC2XFLI.tsx
Boda805: Good point, I think only place possible would be getSupplyCap on error, I just proposed changing return on catch to 1 instead of 0

controtie: IMO it'd be better to address that here, since this now has a dependency on getSupplyCap returning 1. That's just a nit pick though.

controtie (approved):
Look great! Thanks for your effort @Boda805 !

PR #336, opened 2021-08-09
Improve market cap

### LINKED ISSUE #330: BED Page and other Token Stats bugs
**Describe the bug**
Coingecko is slow to roll out the MCAP for BED, and this seems to be a common pattern with them. Alongside this, there seems to be a bug caused by this that prevents the hiding of the supply cap stat, which should not show for simple indices

**Expected behavior**
The BED product page should show an MCAP differently than it currently does. 
By default, the value should be '--'. 
If MCAP is provided by CoinGecko, instead show it. 
Else, a rough estimate of MCAP should be calculated, formatted, and show (currentSupply * CG Price). This also requires a user to connect their wallet. You might have to pass this value in and add it to ProductTokenStatsProps (this will cascade to a few changes on each product page)

For supply cap: @ src/components/ProductPage/ProductTokenStats.tsx:30
image
Move `supplyCap ?` to line 39 below, following the pattern on streamingFee (ln 32)
image

**Screenshots**
image

**Additional context**
Supply data is coming from the token contract, so it will require a provider.

rootulp:
Fixes #330

If a market cap is provided, display it.
If total supply is provided, approximate market cap as supply * price. This may be useful for future product launches where CoinGecko doesn't provide market cap on day 1.
If market cap and supply aren't provided, display "--"

## Screenshots for DPI

Scenario | Screenshot
--- | ---
If market cap is provided | Screen Shot 2021-08-09 at 6 10 05 PM
If total supply is provided | Screen Shot 2021-08-09 at 6 09 42 PM
Neither are provided | Screen Shot 2021-08-09 at 6 09 22 PM

Tagging @0xModene

### DIFF
--- a/src/components/ProductPage/ProductTokenStats.tsx
+++ b/src/components/ProductPage/ProductTokenStats.tsx
@@ -22,12 +22,24 @@ const ProductTokenStats: React.FC<ProductTokenStatsProps> = ({
   fees,
   supplyCap,
   currentSupply,
+  netAssetValue,
 }) => {
   const formatMetric = (metricValue: number) =>
     numeral(metricValue).format('0.00a').toString().toUpperCase()
 
   const formattedSupplyCap = () =>
     supplyCap ? numeral(supplyCap?.toString() || '0').format('0,0') : '--'
+
+  const formattedMarketCap = () => {
+    if (latestMarketCap) {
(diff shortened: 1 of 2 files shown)


controtie (approved):
LGTM

PR #337, opened 2021-08-09
Update browserslist

rootulp:
Fixes a warning observed when running `yarn install`. Add a section
about it to the README.

## Testing
I checked that the app still works locally after updating

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "apollo-client": "^2.6.10",
     "axios": "^0.21.1",
     "bignumber.js": "^9.0.1",
-    "caniuse-lite": "^1.0.30001245",
     "canvas-confetti": "^1.3.0",
     "google-spreadsheet": "^3.1.15",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -16,7 +16,6 @@
     "apollo-client": "^2.6.10",
     "axios": "^0.21.1",
     "bignumber.js": "^9.0.1",
-    "caniuse-lite": "^1.0.30001245",
```
rootulp, line 19: `npx browserslist --update-db` removed this after running. In the logs I see:

```
Cleaning package.json dependencies from caniuse-lite
$ yarn remove -W caniuse-lite
```

According to https://github.com/browserslist/browserslist#browsers-data-updating this seems like an instance of `caniuse-lite` deduplication

controtie (approved):
Great stuff @rootulp ! Awesome to see you jumping right in 🙂
Welcome to the Coop!

PR #338, opened 2021-08-11
Remove Surface fill prop

rootulp:
- Remove the `fill` prop from the `<Surface>` component because it
  causes errors of the form
  https://github.com/clintonbembryjr/react-neu/issues/4
- The page looks identical without this prop


controtie:
> On a related note, we may consider moving off of the https://www.npmjs.com/package/react-neu dependency because it doesn't see much usage (26 downloads this week, last release 10 months ago).

100%. This is something we inherited as we forked our repo from YAM's UI.

controtie (approved):
LGTM

PR #376, opened 2021-09-08
Fix Cypress tests

rootulp:
### Description
- Fix Cypress tests so we can get signal from the Github workflow that auto-runs Cypress tests for us.

### Note for maintainers
- Sorry for the large PR. I couldn't find a sensible way to split this up.
- Cypress [best practices](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) guide recommends using `data-*` attributes over dynamic class ids because the latter is brittle. We should encourage contributors to use `data-cy` as a mechanism to find elements on the page from Cypress tests.

### Screenshots

Before | After
--- | ---
Screen Shot 2021-09-08 at 1 52 33 AM | Screen Shot 2021-09-08 at 1 31 50 AM


cypress/integration/liquiditymining.spec.js
```diff
@@ -22,105 +22,91 @@ describe('Liquidity Mining', () => {
...
     it('should show capital in farms', () => {
       cy.get(
-        ':nth-child(1) > .sc-jSFjdj > .SFMFb > .ctlBxL > .etUwLo > .QxpiM > .sc-cTJkRt'
-      ).should('be.visible')
-      cy.get(
-        ':nth-child(1) > .sc-jSFjdj > .SFMFb > .ctlBxL > .etUwLo > .fgTdQE > .sc-efHYUO'
-      ).should('not.be.empty')
-      cy.get(
-        ':nth-child(1) > .sc-jSFjdj > .SFMFb > .ctlBxL > .etUwLo > .fgTdQE > .sc-iBzEeX'
+        ':nth-child(1) > .sc-jSFjdj > .SFMFb'
```
rootulp, line 25: I couldn't get unique `data-cy` ids attached to elements in the liquidity mining page but other pages should have more IDs to target.

cypress/integration/resources/vote.spec.js
```diff
@@ -1,33 +0,0 @@
-/// <reference types="cypress" />
-
-describe('Vote', () => {
```
rootulp, line 3: We link to Snapshot now so this test file can go away

src/utils/cypressUtils.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * formatCypressId returns a string that can be used as a Cypress data ID
+ * selector (e.g. data-cy)
+ */
+export function formatCypressId(input: string): string {
+    // TODO: add unit tests when this package enables Jest
```
rootulp, line 6: Are maintainers open to me adding Jest to this package?

rootulp:
Tagging @0xModene @controtie as reviewers

src/utils/cypressUtils.ts
0xModene: sure

cypress/integration/liquiditymining.spec.js
0xModene: to resolve this, add 2 seconds to line 5 like `cy.wait(5000)`. This page is slow, so if you wait a couple more seconds here then it should load properly

src/utils/cypressUtils.ts
rootulp: Disregard, I thought Jest wasn't set up because I didn't see a jest config file. However, create-react-app includes this out of the box: https://create-react-app.dev/docs/running-tests

I'll push a commit to this PR with unit tests for this shortly

cypress/integration/liquiditymining.spec.js
rootulp: Thanks for the tip! I just tried this with no luck. 

Screen Shot 2021-09-08 at 8 01 20 PM

It turns out the page loads fine in 3 seconds but any `data-cy` attributes that I attach to elements in [Treasury](https://github.com/SetProtocol/index-ui/blob/master/src/views/Farm/components/Treasury.tsx) don't end up persisted. One observation I've made is that all the components that I'm trying to attach ids to are react-neu [Card](https://github.com/clintonbembryjr/react-neu/tree/master/src/components/Card) elements but I'm not sure why that could be throwing this off

0xModene (changes requested):
This is almost done. All looks good, but can you undo the xdescribe on the Homepage tests as well?

rootulp:
> This is almost done. All looks good, but can you undo the xdescribe on the Homepage tests as well?

Sure, can do! Will follow up with another commit shortly

cypress/integration/liquiditymining.spec.js
rootulp: Confirmed Card components are also broken on the homepage. https://github.com/clintonbembryjr/react-neu/issues/5

rootulp:
Homepage Cypress tests pass

Screen Shot 2021-09-08 at 8 40 05 PM

Although the tests on the homepage and liquidity mining page are super brittle. We need to fix https://github.com/clintonbembryjr/react-neu/issues/5 before we can create more robust selectors for elements on these pages

0xModene (approved):
Lgtm!

PR #380, opened 2021-09-13
Read streaming fees from on-chain via set.js

rootulp:
## Description
Previously all streaming fees were hard-coded in `productTokens.ts` [example](https://github.com/rootulp/index-ui/blob/b8693c00449054187c6466acbb780423ff53b4e0/src/constants/productTokens.ts#L23-L23). This change reads streaming fees for all products from on-chain. This helps us future-proof in-case streaming fees for existing products are modified. This also sets up infrastructure to read other data from [set.js](https://github.com/SetProtocol/set.js) which will be helpful for future PRs.

## Motivation
[Google doc](https://docs.google.com/document/d/1LXgI6q_YsekAeexw2f6eZBnVdnlrUSwjqYzvYtR0fCw)

## Smoke test for reviewers
1. Pull this PR locally
1. `yarn start`
1. Connect Metamask wallet
1. Observe streaming fees for all products are the same as before this change


rootulp:
In testing I observed no difference between the Token sets server quantity and the on-chain quantity:

src/contexts/Set/getSetDetails.ts
```diff
@@ -0,0 +1,34 @@
...
+    navIssuanceModuleAddress: navIssuanceModuleAddress,
+    protocolViewerAddress: protocolViewerAddress,
+    setTokenCreatorAddress: setTokenCreatorAddress,
+    streamingFeeModuleAddress: streamingFeeModuleAddress,
+    tradeModuleAddress: tradeModuleAddress,
+    governanceModuleAddress: governanceModuleAddress,
+    debtIssuanceModuleAddress: debtIssuanceModuleAddress,
+  });
+
+  return set.setToken.batchFetchSetDetailsAsync([productAddress], []);
```
controtie, line 33: I believe you can pass in modules as the second argument here to fetch related details. (Not that you would know that given our documentation...)

Anyways, here's a snippet from our tokensets repo:
```
    const moduleAddresses = [
      isPolygon
        ? process.env.POLYGON_SETJS_BASIC_ISSUANCE_MODULE
        : process.env.SETJS_BASIC_ISSUANCE_MODULE,
      isPolygon
        ? process.env.POLYGON_SETJS_STREAMING_FEE_MODULE
        : process.env.SETJS_STREAMING_FEE_MODULE,
      isPolygon ? process.env.POLYGON_SETJS_TRADE_MODULE : process.env.SETJS_TRADE_MODULE,
      isPolygon
        ? process.env.POLYGON_SETJS_DEBT_ISSUANCE_MODULE
        : process.env.SETJS_DEBT_ISSUANCE_MODULE,
    ];

    const setDetailsList: ISetDetails[] = await setJS.setToken.batchFetchSetDetailsAsync(
      addresses,
      moduleAddresses,
      caller,
    );
```

controtie (changes requested):
Thanks for the super details comments as usual! :D 

I think we definitely want to read the quantities on chain here.
1. Set _eventually_ wants to deprecate its APIs and go pretty decentralized
2. Set runs a scraper bot in the background to update it's DB with set positions. If we're doing something like issuance/redemption we need to be pretty sure the position's we're reading are accurate.

I added a comment re: fetching streaming fee quantities on-chain. This info probably isn't included in the docs. Sorry about that!

rootulp:
WIP. I'm observing that streaming fees for DPI and MVI appear to be 0 when read from on-chain. I need to investigate if set.js is actually fetching these values

Screen Shot 2021-09-15 at 9 46 15 PM

src/constants/productTokens.ts
```diff
@@ -9,7 +9,7 @@ export interface ProductToken {
   image: string
   coingeckoId: string
   tokensetsId: string
-  fees: any | undefined
+  fees: { streamingFee: string } | undefined
```
rootulp, line 12: Proposal for a refactor that is out of scope for this PR:
Modify `IndexToken` to not be of type `ProductToken` but instead pick a subset of props (all props excluding `fees`) b/c all other product tokens have `fees`. Then we can make this field:

```typescript
  fees: { streamingFee: string }
```

src/contexts/Set/getSetDetails.ts
rootulp: Thanks for sharing! I'll use `moduleAddresses` when I fetch set details in a future PR

src/constants/productTokens.ts
0xModene: I like this idea. There are some other small bits with how it's currently set up I would also like to change. We may need to discuss this soon.

0xModene (changes requested):
only small nits: in many of the files you're importing you're importing `'../../constants~'` but you don't have to. Can you double check all your files and change anything with a leading `../` to just 'constants' or 'utils' etc?


this looks great otherwise, looking forward to getting this in

src/hooks/useStreamingFee.ts
```diff
@@ -0,0 +1,9 @@
+import { useContext } from 'react'
+import { DpiIndexComponentsContext } from 'contexts/DpiIndexComponents'
```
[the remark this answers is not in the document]
rootulp: Good catch, this looks residual from an earlier commit on this PR. Will remove

src/contexts/StreamingFee/StreamingFeeContext.ts
```diff
@@ -0,0 +1,20 @@
+import { createContext } from 'react'
+import { BedIndex, Bitcoin2xFlexibleLeverageIndex, DefiPulseIndex, Ethereum2xFlexibleLeverageIndex, MetaverseIndex } from "../../constants/productTokens"
```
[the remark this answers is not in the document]
rootulp: Out of curiosity, why are the absolute imports preferred over relative imports?

0xModene (approved):
lgtm!

src/contexts/StreamingFee/StreamingFeeContext.ts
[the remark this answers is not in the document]
controtie: It's mostly that, plus not having to know which directory your file is in and a bit easier on the eyes. Open to changing it though if there are benefits of going relative.

src/contexts/StreamingFee/StreamingFeeProvider.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useState, useEffect } from 'react'
+import { getStreamingFees } from "../../utils/setjsApi"
+import { provider } from "web3-core"
+import useWallet from "../../hooks/useWallet"
+import { bedTokenAddress, btc2xfliTokenAddress, dpiTokenAddress, eth2xfliTokenAddress, mviTokenAddress } from "constants/ethContractAddresses"
+import { convertToPercentage } from "../../utils/ethersBigNumber"
+import StreamingFeeContext from "./StreamingFeeContext"
```
controtie, line 7: Per Modene's comment cleaning up these imports to be absolute + ordering (external packages at the top of the file)

src/views/MVI/MVI.tsx
```diff
@@ -2,10 +2,11 @@ import React, { useEffect } from 'react'
 import useMviTokenMarketData from 'hooks/useMviTokenMarketData'
 import useMviComponents from 'hooks/useMviComponents'
 import useBalances from 'hooks/useBalances'
-import { MetaverseIndex } from 'constants/productTokens'
+import { MetaverseIndex, ProductToken } from 'constants/productTokens'
 import ProductDataUI, {
   TokenDataProps,
 } from 'components/ProductPage/ProductDataUI'
+import useStreamingFee from "../../hooks/useStreamingFee"
```
controtie, line 9: Lets make these absolute as well

controtie (changes requested):
My apologies for taking so long to review these changes. I notice the PR changed from reading token allocations of products on-chain to reading streaming fees on-chain. Is this meant to be a small bite on the way to reading token allocations on-chain?

PR looks mostly good besides the minor import clean up!

src/contexts/StreamingFee/StreamingFeeProvider.tsx
rootulp: Can do. With regards to ordering, if we're going to to enforce this manually (i.e. without the help of a linter) then I propose we draft a style guide and link it from the README of this package. I can take a first stab at this with this rule if we feel strongly about it.

On the other hand, I like projects where `lint --fix` automatically formats the code according to all team style guide code conventions. Prettier / ESLint are a solid step towards this. With regards to import ordering, we could enforce something like https://eslint.org/docs/rules/sort-imports

controtie: I'm all for an auto linter/prettifier!

controtie (approved):
Looks good!
Noting that we should make a PR to display mint/redeem fees on BTC/ETH2x-FLI in the future. Not sure those are exposed by set.js atm though.

rootulp:
> Noting that we should make a PR to display mint/redeem fees on BTC/ETH2x-FLI in the future. Not sure those are exposed by set.js atm though.

Added to [google doc](https://docs.google.com/document/d/1LXgI6q_YsekAeexw2f6eZBnVdnlrUSwjqYzvYtR0fCw)

PR #382, opened 2021-09-15
Add headers to Cypress homepage test

rootulp:
Attempt to fix the Cypress homepage test which is failing due to:
```error
  1) Homepage
       HomeHeader
         "before all" hook for "should exist":
     CypressError: `cy.visit()` failed trying to load:

http://localhost:3000/

We attempted to make an http request to this URL but the request failed without a response.

We received this error at the network level:

  > Error: ESOCKETTIMEDOUT
```

Fix is based on comments in https://github.com/cypress-io/cypress/issues/943

### DIFF
--- a/cypress/integration/homepage.spec.js
+++ b/cypress/integration/homepage.spec.js
@@ -3,7 +3,11 @@
 describe('Homepage', () => {
   before(() => {
     cy.wait(3000)
-    cy.visit('http://localhost:3000/')
+    cy.visit('http://localhost:3000/', {
+      headers: {
+        "Accept-Encoding": "gzip, deflate"
+      }
+    })
   })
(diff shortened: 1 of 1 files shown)


cypress/integration/homepage.spec.js
```diff
@@ -3,7 +3,11 @@
 describe('Homepage', () => {
   before(() => {
     cy.wait(3000)
-    cy.visit('http://localhost:3000/')
+    cy.visit('http://localhost:3000/', {
+      headers: {
+        "Accept-Encoding": "gzip, deflate"
```
0xModene, line 8: 🔥

0xModene (approved):
lgtm

PR #383, opened 2021-09-15
Trigger Cypress tests on external PRs

rootulp:
Needed to test https://github.com/SetProtocol/index-ui/pull/382
Based on https://docs.github.com/en/actions/reference/events-that-trigger-workflows


rootulp:
oh lol wat? This change got applied and so the GH workflow actually runs on this PR?! super cool

Screen Shot 2021-09-15 at 12 00 40 AM

0xModene (approved):

PR #394, opened 2021-09-21
Balances to use Wei 

0xModene:
# **Pull Request Template**

###### *Select a category that relates to the content of your pull request (choose all that apply).*

- [ ] Bug Fix
- [ ] Content
- [ ] New Feature
- [ ] Documentation
- [x] Code Refactoring

&nbsp;

## **Summary of Changes**

Balances will now be in wei, and only displayed otherwise. 

Incidentally, also resolved 'NaN' display issue on DPI in the WalletModal code

&nbsp;

Screen Shot 2021-09-22 at 11 43 53 AM
Screen Shot 2021-09-22 at 11 44 09 AM
Screen Shot 2021-09-22 at 11 44 19 AM
Screen Shot 2021-09-22 at 11 44 24 AM

&nbsp;

###### *By submitting this pull request, you are confirming the following to be true:*


src/contexts/Balances/Provider.tsx
```diff
@@ -25,43 +25,41 @@ import {
 } from 'constants/ethContractAddresses'
 
 const Provider: React.FC = ({ children }) => {
```
controtie, line 27: My sense is these should _all_ be `BigNumber` or `strings`. Having a mixture of BN & strings adds unnecessary complexity IMO. I'm leaning towards keeping things `BigNumber`, just to make sure no one is trying to do any math on these that does not involve Big Numbers.

controtie: I think this will also make it match the SetJS integration once we get there.

controtie (changes requested):
IMO we should be using `BigNumber` for storing the user balances. I'm willing to take pushback on this though.

My thinking: there isn't really a case where we want to use WEI in string format. Any calculations/display we want to do will involve `BigNumber` anyways.
We also don't want to encourage doing any math (or much of anything, really) on the raw string versions of user balances.

Additionally, I think this was started before the DATA launch? Looks like it could use a rebase.

PR #395, opened 2021-09-22
Increase response timeout for homepage test

rootulp:
Motivation in https://github.com/cypress-io/cypress/issues/7062#issuecomment-752458819
Attempt to fix:
```
  1) Homepage
       HomeHeader
         "before all" hook for "should exist":
     CypressError: `cy.visit()` failed trying to load:
```


rootulp:
Failed Cypress with a different error:

```
Error: ENOSPC: System limit for number of file watchers reached, watch 'cypress/integration'
```

which led me to https://github.com/cypress-io/cypress/issues/6960

PR #398, opened 2021-09-23
Read token current supply from set.js

rootulp:
### Description
- Read token current supply from set.js
- Since we previously fetched current supply from on-chain by inspecting the ERC20 token contracts, there is no user-visible change with this PR.
- Motivation in this [google doc](https://docs.google.com/document/d/1LXgI6q_YsekAeexw2f6eZBnVdnlrUSwjqYzvYtR0fCw). This PR fulfills goal 2

### Testing
- `yarn start` and verified that current supply for all products matches what is already displayed

### Screenshots for DPI
Before | After
--- | ---
Screen Shot 2021-09-22 at 10 32 36 PM | Screen Shot 2021-09-22 at 10 32 18 PM


src/contexts/TokenSupply/TokenSupplyProvider.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [bedTotalSupply, setBedTotalSupply] = useState<BigNumber>()
+  const [eth2xfliTotalSupply, setEth2xfliTotalSupply] = useState<BigNumber>()
+  const [btc2xfliTotalSupply, setBtc2xfliTotalSupply] = useState<BigNumber>()
+
+  useEffect(() => {
+    if (ethereum && dpiTokenAddress && mviTokenAddress && bedTokenAddress && eth2xfliTokenAddress && btc2xfliTokenAddress) {
+      getTokenSupply(ethereum, [dpiTokenAddress, mviTokenAddress, bedTokenAddress, eth2xfliTokenAddress, btc2xfliTokenAddress])
+        .then(result => {
+          const [dpiResult, mviResult, bedResult, eth2xFliResult, btc2xFliResult] = result
+          setDpiTotalSupply(new BigNumber(dpiResult.totalSupply.toString()).dividedBy(new BigNumber(10).pow(18)))
```
rootulp, line 23: even though Set returns an ethers `BigNumber`, index-ui uses `BigNumber` from bignumber.js so I elected to convert types here

src/views/BED/BED.tsx
```diff
@@ -30,7 +32,7 @@ const BedProductPage = (props: { title: string }) => {
     token: token,
     components: components,
     balance: bedBalance,
-    currentSupply: bedTotalSupply,
+    currentSupply: bedTotalSupply ?? bedTotalSupplyFromBalances,
```
rootulp, line 35: here we fallback to the existing supply fetched from `useBalances()` in case the request to set.js fails

`??` is the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)

src/contexts/TokenSupply/TokenSupplyProvider.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useState, useEffect } from 'react'
+import { provider } from "web3-core"
+
+import { getTokenSupply } from "utils/setjsApi"
+import useWallet from "hooks/useWallet"
+import { bedTokenAddress, btc2xfliTokenAddress, dpiTokenAddress, eth2xfliTokenAddress, mviTokenAddress } from "constants/ethContractAddresses"
+import TokenSupplyContext from "./TokenSupplyContext"
+import BigNumber from "bignumber.js"
```
controtie, line 8: Move to top of file

src/contexts/TokenSupply/TokenSupplyProvider.tsx
controtie: Makes sense! ethers' `BigNumber` can be a bit of a pain to use.

src/utils/setjsApi.ts
```diff
@@ -1,9 +1,27 @@
 import { provider } from 'web3-core'
 import Set from "set.js"
-import { StreamingFeeInfo } from "set.js/dist/types/src/types";
+import { SetDetails, StreamingFeeInfo } from "set.js/dist/types/src/types";
 
 import { basicIssuanceModuleAddress, controllerAddress, debtIssuanceModuleAddress, navIssuanceModuleAddress, protocolViewerAddress, streamingFeeModuleAddress, setTokenCreatorAddress, tradeModuleAddress, masterOracleAddress, governanceModuleAddress } from "constants/ethContractAddresses";
 
+export async function getTokenSupply(web3Provider: provider, productAddresses: string[]): Promise<SetDetails[]> {
```
controtie, line 7: This is interesting. Given we're fetching more than just Set Details and not just Token Supply here, does it make sense to rename this?

Or maybe it could make sense to alter `TokenSupplyProvider` to be a general `SetDetailsProvider`? And add a selector for the `tokenSupply` part of the `setDetails` response.

If we want to do that in a follow up PR instead of here that's fine too.

src/views/BTC2XFLI/BTC2XFLI.tsx
```diff
@@ -13,6 +13,7 @@ import ProductDataUI, {
 import useWallet from 'hooks/useWallet'
 import BigNumber from 'utils/bignumber'
 import useStreamingFee from "hooks/useStreamingFee"
+import useTokenSupply from "../../hooks/useTokenSupply"
```
controtie, line 16: absolute path

src/views/ETH2XFLI/ETH2XFLI.tsx
```diff
@@ -13,6 +13,7 @@ import ProductDataUI, {
 import useWallet from 'hooks/useWallet'
 import BigNumber from 'utils/bignumber'
 import useStreamingFee from "hooks/useStreamingFee"
+import useTokenSupply from "../../hooks/useTokenSupply"
```
controtie, line 16: absolute path

src/views/BED/BED.tsx
controtie: sweet, lets use this more @0xModene

controtie (changes requested):
Looks mostly good. I think it makes sense to convert TokenSupplyProvider to SetDetailsProvider since we'll want other stuff from the `batchFetchSetDetails` response anyways.

Also DATA just launched, can you add an entry for that as well? :D

src/contexts/TokenSupply/TokenSupplyProvider.tsx
rootulp: Good to know. I've only use bignumber.js and Ethers BigNumber sparingly so I'm not sure which one is preferable for this repo to use. Seems safe to keep bignumber.js until we have convincing evidence to switch off

src/utils/setjsApi.ts
rootulp: I'm down to do this rename in a future PR when I tackle reading product components via set.js

rootulp (commented):
FYI current supply for DATA before and after this change is 0

Screen Shot 2021-09-28 at 11 22 27 PM

0xModene:
> FYI current supply for DATA before and after this change is 0
> 
> Screen Shot 2021-09-28 at 11 22 27 PM

This will be the case for roughly another week or so with CoinGecko data. It's still the case with on-chain data?

0xModene:
Accidental "Close With Comment", sorry

rootulp:
> > FYI current supply for DATA before and after this change is 0
> > Screen Shot 2021-09-28 at 11 22 27 PM
> 
> This will be the case for roughly another week or so with CoinGecko data. It's still the case with on-chain data?

I'm surprised it's 0 with on-chain data but can investigate and tackle in a separate PR given behavior isn't impacted by this PR.

src/contexts/TokenSupply/TokenSupplyProvider.tsx
controtie: Yeah, so basically ethers` BigNumber has a bunch of safety things built in (can't use floats) to ensure everything's compatible with the EVM (which doesn't allow floats). Way too onerous IMO though, I prefer converting to local BigNumber

controtie:
> > > FYI current supply for DATA before and after this change is 0
> > > Screen Shot 2021-09-28 at 11 22 27 PM
> > 
> > 
> > This will be the case for roughly another week or so with CoinGecko data. It's still the case with on-chain data?
> 
> I'm surprised it's 0 with on-chain data but can investigate and tackle in a separate PR given user-visible behavior isn't impacted by this PR.

It's actually showing up for me. Make sure you have the address for DATA in your .env file?

controtie (approved):
LGTM 🏄‍♀️

rootulp:
> It's actually showing up for me. Make sure you have the address for DATA in your .env file?

🤦‍♂️ that was it. Works now:

Screen Shot 2021-09-29 at 5 40 38 PM

which matches [etherscan](https://etherscan.io/token/0x33d63ba1e57e54779f7ddaeaa7109349344cf5f1)

Screen Shot 2021-09-29 at 5 41 41 PM

PR #404, opened 2021-09-29
Fix DATA img alt warning

rootulp:
Fixes

Screen Shot 2021-09-28 at 11 27 24 PM


0xModene (approved):
what a goofy error, thanks for catching this and fixing it

PR #422, opened 2021-10-12
Read set component quantities from on-chain

rootulp:
## **Type of Change**

- [x] Code Refactoring

## **Summary of Changes**

The goal was to replace requests to the Set backend by reading product underlying tokens and their allocations from on-chain. See [google doc](https://docs.google.com/document/d/1LXgI6q_YsekAeexw2f6eZBnVdnlrUSwjqYzvYtR0fCw/edit#) for more context.

I observed sporadic 429s from CoinGecko during development. I think a paid tier CoinGecko API key would prevent indexcoop.com from observing similar throttles in production. I haven't observed these throttles since batching requests correctly so this may not strictly be necessary.

Note: I removed the 24hr change column
Product | Before | After
--- | --- | ---
DPI | Screen Shot 2021-11-12 at 11 48 29 PM | Screen Shot 2021-11-12 at 11 48 09 PM
MVI | Screen Shot 2021-11-12 at 11 50 12 PM | Screen Shot 2021-11-12 at 11 49 59 PM


0xModene:
> * Fetch all index components in one request

I'm looking into doing this for a lot of other things as well. I think we should combine the token calls by categories, not by token. Will require other sizeable changes, but will be cleaner and help get lessen the provider shelf in `App.tsx`

rootulp:
@0xModene or @controtie do ya'll have insight on how `percentOfSet` is calculated on Set's server side? My naive idea on how to do this is get the CoinGecko price of each position and multiply by quantity of each position.

If we got this route, I think we want to make a batch CoinGecko request to make this performant.

0xModene:
> @0xModene or @controtie do ya'll have insight on how `percentOfSet` is calculated on Set's server side? My naive idea on how to do this is get the CoinGecko price of each position and multiply by quantity of each position.
> 
> If we got this route, I think we want to make a batch CoinGecko request to make this performant.

This data is all [on chain](https://etherscan.io/token/0x1494ca1f11d487c2bbe4543e90080aeba4ba3c2b#readContract), and I believe it's just simple math

@ncitron may have a better idea

ncitron:
Fetching the amount of an underlying component per set token is relatively easy. This can be done through a call to `setToken.getDefaultPositionRealUnits(underlyingComponentAddress)`. From there you can just fetch the prices from coingecko. One thing to be careful about though is the FLI tokens. For FLI, each token has two components: an equity amount of compound deposit tokens (cTokens), and a negative debt amount of USDC.

rootulp:
Hmm the fleek/build preview isn't working for me: https://bafybeic5zrh7bghwbr7g5fcyqhkrscanqbc7scm7wrpzuqp3u2bhfb56ny.on.fleek.co/

I'm getting rate limited by CoinGecko locally so was hoping to test the most recent [WIP] commit via this preview

rootulp:
## To do
- [ ] Do we want to sort the components by percent of Set in descending order (?)
- [ ] Fix NaN for WAXE on Metaverse page
- [ ] Fix NaNs on Eth2xFli page
 - [ ] Fix NaNs on Btc2xFli page
- [ ] DPI logo doesn't display for me

## FLUPs (out of scope for this PR)
- [ ] Consider using Uniswap on-chain price for tokens rather than CoinGecko

rootulp:
oof

Screen Shot 2021-11-10 at 9 38 07 PM

on http://localhost:3000/ethfli

rootulp:
FYI @controtie this will be blocked until we get a CoinGecko pro plan API key

src/contexts/SetComponents/SetComponentsProvider.tsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function getPositionPrices(setDetails: SetDetails): Promise<any> {
+  const componentAddresses = setDetails.positions.map(p => p.component)
+  return fetch(`https://api.coingecko.com/api/v3/simple/token_price/${ASSET_PLATFORM}?vs_currencies=${VS_CURRENCY}&contract_addresses=${componentAddresses}`)
+    .then(response => response.json())
+    .catch(e => console.error(e))
+}
+
+// HACKHACK is there a safer way to fetch decimal places for these components?
```
rootulp, line 134: Do reviewers know of a safer way to determine this? I naively assumed all component quantities would have 18 significant figures but I had to divide by 4, 6, 8, or 18 to get the component quantities for products like Eth2xFli to match up with what is visible on https://www.tokensets.com/explore

rootulp:
@0xModene @controtie ready for review. Sorry for my delay on this PR.

0xModene:
@rootulp you mention batching requests correctly, and theres a TON of that we could do across the app. We may need to look more into that

src/contexts/SetComponents/SetComponentsProvider.tsx
0xModene: That's actually a read only function on the contract called `decimals()` with no params, returns a BN. 

This will need to be checked for sure. MVI has a couple of tokens with different significant digits (WHALE comes to mind)

rootulp:
> @rootulp you mention batching requests correctly, and theres a TON of that we could do across the app. We may need to look more into that in another story

Sounds good. During development, I noticed that we were performing two network requests per product to fetch market chart data. This seems potentially wasteful. E.g.
1. https://api.coingecko.com/api/v3/coins/data-economy-index/market_chart?vs_currency=usd&days=90
2. https://api.coingecko.com/api/v3/coins/bankless-bed-index/market_chart?vs_currency=usd&days=max&interval=daily

Upon further inspection, it looks like the `days=max` query returns data at a lower granularity than `days=90`. From [CoinGecko api docs](https://www.coingecko.com/en/api/documentation)

> Minutely data will be used for duration within 1 day, Hourly data will be used for duration between 1 day and 90 days, Daily data will be used for duration above 90 days.

We likely want the hourly data if we continue supporting the one day and one week options on the market chart.

Screen Shot 2021-11-13 at 12 52 13 PM

@0xModene are there other non-batched CoinGecko requests you are interested in? Down to tackle in a separate GH issue

src/contexts/SetComponents/SetComponentsProvider.tsx
rootulp: yep WHALE and WAXE are definitely oddities so I special cased them in this function. Thanks for sharing the `decimals()` function. I manually verified the tokens that I special cased in the function and they all line up.

Symbol | decimals | contract
--- | --- | ---
WHALE | 4 | https://etherscan.io/address/0x9355372396e3F6daF13359B7b607a3374cc638e0#readContract
USDC | 6 | https://etherscan.io/address/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48#readProxyContract
WBTC | 8 | https://etherscan.io/address/0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599#readContract
WAXE | 8 | https://etherscan.io/address/0x7a2Bc711E19ba6aff6cE8246C546E8c4B4944DFD#readProxyContract
CETH | 8 | https://etherscan.io/address/0x4Ddc2D193948926D02f9B1fE9e1daa0718270ED5#readContract
CWBTC | 8 | https://etherscan.io/token/0xC11b1268C1A384e55C48c2391d8d480264A3A7F4#readContract

Considering this PR diff is huge, I'm inclined to merge this as-is and tackle reading `decimals()` from on-chain for all token contracts in a fast-follow PR if that's okay with reviewers.

rootulp: Created https://github.com/SetProtocol/index-ui/issues/499

0xModene:
@rootulp actually those are coming for different reasons. hourly data can be had up to 90d, but not beyond through CG. The second call is got the 12m+ data charts

src/components/ProductPage/ProductDataUI.tsx
```diff
@@ -22,6 +21,7 @@ import {
   ProductToken,
 } from 'constants/productTokens'
 import BigNumber from 'utils/bignumber'
+import { SetComponent } from "../../contexts/SetComponents/SetComponent"
```
controtie, line 24: Prefer absolute import

src/contexts/SetComponents/SetComponent.ts
```diff
@@ -0,0 +1,56 @@
...
+     * @example "UNI"
+     */
+    symbol: string
+
+    /**
+     * Total price of this component. This is equivalant to quantity of
+     * component * price of component.
+     */
+    totalPriceUsd: string
+}
```
controtie, line 56: nice! :D

src/contexts/SetComponents/SetComponentsProvider.tsx
```diff
@@ -0,0 +1,146 @@
...
+  )
+}
+
+async function convertPositionToSetComponent(position: Position, tokenList: Token[], componentPriceUsd: number, setPriceUsd: number): Promise<SetComponent> {
+  const token = getTokenForPosition(tokenList, position)
+  const quantity = new BigNumber(position.unit.toString()).div(new BigNumber(10).pow(getDecimalPlaces(token.symbol)));
+  const totalPriceUsd = quantity.multipliedBy(componentPriceUsd);
+  const percentOfSet = totalPriceUsd.dividedBy(setPriceUsd).multipliedBy(100)
+
+  if (token.symbol.toLowerCase() === "ceth" || token.symbol.toLowerCase() === 'usdc') {
```
controtie, line 93: The way you can tell if a component is a debt position is by looking at the `positionState` field.
Here's an excerpt from ETH2x-FLI
image
If `positionState === 0` it's a normal component
If `positionState === 1` it's a held in an external contract (likely a debt position)

controtie (approved):
Hey this is a really great first draft @rootulp !
Left some comments re: pulling some data but overall it's really easy to read for something so complicated.

One more ask would be putting in a placeholder text when the user's wallet isn't connected under asset allocation.

Maybe something like `Connect Your Metamask` under the allocations section?

controtie (changes requested):
Ping me if you have any questions!

0xModene:
@controtie I'll look today for a way to lint external/internal imports

src/contexts/SetComponents/SetComponentsProvider.tsx
rootulp: Thanks! These console.logs are from debugging. I'll remove them!

src/contexts/SetComponents/SetComponentsProvider.tsx
rootulp: That makes this _way_ easier than I expected. Will fix in this PR

rootulp:
> Maybe something like `Connect Your Metamask` under the allocations section?

Screen Shot 2021-11-15 at 6 32 24 PM

How's this? I intentionally didn't say Metamask b/c indexcoop.com supports connecting with other wallets. `Unlock wallet` matches the text in the top right button.

src/App.tsx
```diff
@@ -122,10 +120,14 @@ const App: React.FC = () => {
...
               <HowToBuy title={'Index - How to Buy'} />
             </Route>
-            <Route exact path='/join'
+            <Route
+              exact
+              path='/join'
               render={() => (window.location.href = discordLink)}
             />
-            <Route exact path='/discord'
+            <Route
```
rootulp, line 128: FYI I don't think Prettier actually runs on this codebase b/c this diff was applied when I used VSCode `CMD + P` > Format Document With... > Prettier

I've seen a handful of instances where double quotes are used when single quotes are specified:

https://github.com/SetProtocol/index-ui/blob/master/.prettierrc.json#L3

src/contexts/SetComponents/SetComponentsProvider.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+async function convertPositionToSetComponent(
+  position: Position,
+  tokenList: Token[],
+  componentPriceUsd: number,
+  setPriceUsd: number
+): Promise<SetComponent> {
+  const token = getTokenForPosition(tokenList, position)
+  const quantity = new BigNumber(position.unit.toString()).div(
+    new BigNumber(10).pow(token.decimals)
```
rootulp, line 182: using token.decimals resolved https://github.com/SetProtocol/index-ui/issues/499

0xModene (approved):
lgtm!

rootulp:
The last commit changes copy from "Unlock" to "Connect" to match the button in the top right

Screen Shot 2021-11-16 at 11 32 46 AM

controtie:
I'm having trouble getting the set allocations to show up when connecting my ledger.
The error I'm receiving is `An operation that changes the device state is in progress.`

I'm not an expert on our ledger integration. This shows up in my google search:
https://github.com/crosswalk-project/chromium-crosswalk/blob/master/third_party/WebKit/Source/modules/webusb/USBDevice.cpp

Since this change sort of coincides with the v2 deployment of the website (strongly encouraging users to login with the majority of non-logged in info displayed in the wrapper marketing application) I'd like to point this to the v2-website-master branch and merge it now while we investigate this issue in parallel

controtie (approved):

PR #494, opened 2021-11-09
Fixes the big number bug on the MVI card in the Liquidity Mining page

omnifient:
# **Pull Request Template**

###### *Select a category that relates to the content of your pull request (choose all that apply).*

- [x] Bug Fix
- [ ] Content
- [ ] New Feature
- [ ] Documentation
- [ ] Code Refactoring

&nbsp;

## **Summary of Changes**

1. The issue _**seems to be**_ caused by the fact that the contract returns
uint256, but the proxy functions (`getEarnedIndexTokenQuantity` and `getBalance`) return strings.
- changed `getEarnedIndexTokenQuantity` to return `BigNumber`
- created a `getBigNumBalance` which is a copy of `getBalance` but returns
`BigNumber`
- had to call the aforementioned functions separately in the
`Balances/Provider` because of differing return types

2. In addition, the values are in wei, and are not getting converted to ether units.
- calling `fromWei` in `formattedStakedBalance` and `formattedEarnedBalance`
to convert to the correct units

**Note:** I am not sure if 1) is actually needed - just by looking at the screenshot in the issue, it looks like 2) would fix it, but I checked their balances on-chain and they are no where close to what's being shown.

&nbsp;

### **Fixes: ** 
https://index-coop.notion.site/MVI-BigNumber-Big-4436f182c6a24530b53f8b76ae379b23

&nbsp;

Note: Could not find a better way to test this other than mocking the return values. 

Test data:
Screenshot from 2021-11-09 04-09-07

After:
Screenshot from 2021-11-09 04-10-15

###### *By submitting this pull request, you are confirming the following to be true:*

### DIFF
--- a/src/contexts/Balances/Provider.tsx
+++ b/src/contexts/Balances/Provider.tsx
@@ -4,7 +4,7 @@ import { provider } from 'web3-core'
 
 import Context from './Context'
 import useWallet from 'hooks/useWallet'
-import { getBalance, getEthBalance } from 'utils/index'
+import { getBalance, getBigNumBalance, getEthBalance } from 'utils/index'
 import { getEarnedIndexTokenQuantity } from 'index-sdk/stake'
 import { getEarnedIndexTokenQuantity as getEarnedFarmTwoBalance } from 'index-sdk/farmTwo'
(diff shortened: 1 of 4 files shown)


0xModene (approved):
lgtm! thanks! worked on local tests as you explained

PR #505, opened 2021-11-16
Display ENS name for connected wallet

rootulp:
## Motivation
ENS airdrop last week 🎉🥳 Let's make indexcoop.com ENS aware

## Description
1. Change "Unlock wallet" to "Connect wallet" because the latter seems more common on Web3 sites
2. Display ENS name for connected wallet (if user has an ENS). If not, fallback to shortened version of their public address.

Scenario | Screenshot
--- | ---
Not connected | ens-not-connected
Connected without ENS | ens-without-ens
Connected with ENS | ens-with-ens

## Bundle size

This introduces a new dependency on [useDapp](https://usedapp.readthedocs.io/en/latest/) which increases our bundle size by approx. (1.97MB - 1.89MB) = .08MB
Before | After
--- | ---
bundle-size-before | bundle-size-after

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "@testing-library/jest-dom": "^5.15.0",
     "@testing-library/react": "^12.1.2",
     "@testing-library/user-event": "^13.5.0",
+    "@usedapp/core": "^0.5.5",
     "apollo-client": "^2.6.10",
     "axios": "^0.24.0",
     "bignumber.js": "^9.0.1",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,6 +5,7 @@ import { ToastContainer, Slide } from 'react-toastify'
 import { ApolloProvider } from '@apollo/client'
 import styled from 'styled-components'
 import 'react-toastify/dist/ReactToastify.css'
+import { ChainId, DAppProvider, Config } from '@usedapp/core'
 
 import MobileMenu from 'components/MobileMenu'
 import TopBar from 'components/TopBar'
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -149,65 +154,67 @@ const Providers: React.FC = ({ children }) => {
...
-                          </MviStakingRewardsProvider>
-                        </FarmingTwoProvider>
-                      </FarmingProvider>
-                    </BalancesProvider>
-                  </ExternalAirdropProvider>
-                </RewardsProvider>
-              </AirdropProvider>
-            </MediaQueryProvider>
-          </ApolloProvider>
+          <DAppProvider config={config}>
```
rootulp, line 157: The only line change here is adding `<DAppProvider config={config}>`. Every line below just gets indented by one tab

src/components/TopBar/components/WalletButton.tsx
```diff
@@ -1,9 +1,9 @@
 import React, { useCallback } from 'react'
 import styled from 'styled-components'
-
 import { Button } from 'react-neu'
-import useWallet from 'hooks/useWallet'
+import { shortenAddress, useLookupAddress } from '@usedapp/core'
 
+import useWallet from 'hooks/useWallet'
```
rootulp, line 6: grouped external / internal imports. Separated by newline

package.json
```diff
@@ -10,6 +10,7 @@
     "@testing-library/jest-dom": "^5.15.0",
     "@testing-library/react": "^12.1.2",
     "@testing-library/user-event": "^13.5.0",
+    "@usedapp/core": "^0.5.5",
```
0xModene, line 13: Yes yes, the march toward ethers begins!
muahaha

src/App.tsx
0xModene: soon we will kill some of these providers

0xModene (approved):
lgtm!
I'll wait for #422 to merge first to be sure of no conflict, since that's a huge change

0xModene:
Actually going to go ahead and merge this in since I have to wait on @controtie to review #422

Issue #514, opened 2021-11-19
Investigate Set component allocations with Ledger


rootulp:
> I'm having trouble getting the set allocations to show up when connecting my ledger.
The error I'm receiving is `An operation that changes the device state is in progress.`

> I'm not an expert on our ledger integration. This shows up in my google search:
https://github.com/crosswalk-project/chromium-crosswalk/blob/master/third_party/WebKit/Source/modules/webusb/USBDevice.cpp

_Originally posted by @controtie in https://github.com/SetProtocol/index-ui/issues/422#issuecomment-970588296_

rootulp:
Still investigating. This looks caused by https://github.com/SetProtocol/index-ui/blob/46e42c0122b02948a8a7ce1de1b5d8186caa49a8/src/utils/setjsApi.ts#L70

index-ui doesn't do proper error handling so we should add a `.catch()` to the function that invokes this method. set.js may also need to be updated to catch the error and propagate a stacktrace.

rootulp:
Relevant https://github.com/LedgerHQ/ledgerjs/issues/607

rootulp:
Created https://github.com/LedgerHQ/ledgerjs/issues/718

PR #517, opened 2021-11-21
Upgrade Ledger dependencies

rootulp:
I did these upgrades to try and reoslve https://github.com/SetProtocol/index-ui/issues/514. They didn't resolve that issue but I think we can upgrade them anyway.

This targets **website-v2-master**. I can duplicate for **master** if requested.


controtie:
Hmm, when I tried pulling this down, it looks like it breaks the buy/sell flow. Can you confirm?

PR #528, opened 2021-11-27
Prettier --write src/**

rootulp:
I followed the steps on https://create-react-app.dev/docs/setting-up-your-editor to try and have prettier run as a pre-commit hook but it doesn't appear to work as expected.

The diff in this PR was caused by:
```bash
./node_modules/.bin/prettier --write "src/**/*.{js,jsx,ts,tsx,json,css,scss,md}"
```

I think we'll want to run `yarn lint:fix` or the above periodically until we have a CI check that ensures PRs aren't merged that don't conform to prettier

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,9 +91,9 @@
     "@types/styled-components": "^5.1.15",
     "@types/web3": "^1.2.2",
     "cypress": "^9.1.0",
-    "husky": "^4.3.0",
+    "husky": "^7.0.4",
     "lint-staged": "^12.1.2",
-    "prettier": "2.5.0",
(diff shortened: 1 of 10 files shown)


0xModene:
> I think we'll want to run `yarn lint:fix` or the above periodically

I think this is fine for the time being, since it's better than what we had before. Perhaps we could run a github workflow to the mix and add it as a check that needs to be passed before a merge?

0xModene (approved):
lgtm

rootulp:
> Perhaps we could run a github workflow to the mix and add it as a check that needs to be passed before a merge?

Yep, completely agree!

PR #555, opened 2021-12-12
Use new INDEX token logo on liquidity mining page

rootulp:
### Motivation
Consistency with usage of INDEX token on rest of this app / CoinGecko

Before | After
--- | ---
index-token-before | index-token-after


src/views/Farm/components/Harvest/Harvest.tsx
```diff
@@ -1,67 +0,0 @@
...
-  )
-}
-
-const StyledIcon = styled.img`
-  height: 58px;
-  text-align: center;
-  min-width: 58px;
-`
-
-export default Harvest
```
rootulp, line 67: This component is unused so I opted to delete it. Can preserve if reviewers feel strongly this residual code should remain in the codebase. We can always resurrect it from git history if needed in the future.

0xModene (approved):
lgtm

Issue #389, opened 2021-09-21
Automatic code style formatting


rootulp:
**Describe the solution you'd like**
- Automatic code style formatting in this package. Likely with [Prettier](https://prettier.io/) + [ESLint](https://eslint.org/)

**Additional context**
- Consider enabling default React, TypeScript rules. Then consider adding https://eslint.org/docs/rules/sort-imports
- Motivation in https://github.com/SetProtocol/index-ui/pull/380#discussion_r713312466

0xModene:
I'm here for this, this would at minimum prevent any whitespace related merge/diff issues. We already have a prettier config. Perhaps we can vote on any new rules in #dev-public?

tytarman:
@0xModene I've been working on an import linter. Just wanted to post here to claim the work I've already done on it. It's almost ready for a PR. Just cleaning up some final details, and it'll be ready next week (hopefully)

0xModene:
cool @tytarman I've assigned this to you then

PR #559, opened 2021-12-15
Let there be sorting!

tytarman:
Sorts imports alphabetically and according to type

# **Pull Request Template**

###### _Select a category that relates to the content of your pull request (choose all that apply)._

- [ ] Bug Fix
- [ ] Content
- [x] New Feature
- [ ] Documentation
- [x] Code Refactoring

&nbsp;

### Overview

I have parsed out an import linter using ESLint that is able to check import statements and sort them alphabetically and according to type. The necessary dependencies are already roped in through the package.json file, and I have spent grueling hours testing it on every single `.ts` or `.tsx` file we have.

The linter automatically ignores the `./node_modules` folder, but I also included an ignore file to be sure that it not only ignores `./node_modules` but `./cypress` and `./public` as well. Opening a file within these folders **WILL** show warnings and sort imports upon a local save, but otherwise, they are ignored.

The linter identifies improper formatting as warnings on the command line, and it is capable of auto-fixing problems whenever the current file is saved (tested in VSC). 

&nbsp;

### I chose to sort imports in the following way:
- `import React from 'react'`

- All other `react` imports

- Anything from the `./node_modules` folder

- Anything from the `./src` folder

- Anything that begins with a `.` or `..` or `/`

&nbsp;

The categories are separated by a blank lines to help distinguish between different types of imports. Duplicates aren't allowed, and everything is sorted both alphabetically and numerically within each group.

&nbsp;

### How to Use It

Apart from it automatically recognizing problems in real time, you can simply run `yarn eslint ./fileName` for the linter to search for sort errors. You can also run `yarn eslint ./fileName --fix`, and the linter will automatically fix things for you in bulk. **Test it out on the `./src` folder yourself. Go ahead. It's fun!**

&nbsp;

### New Dependencies

I have been testing this thing for entirely too long in my spare time (like 2 months), and I finally decided that this configuration is the most simple and straightforward way to address the issue. Consequently, new dependencies needed to be included in order for ESLint to function properly. Namely,`eslint-plugin-simple-import-sort` provided the best results out of probably 10 different configurations that I tried, but it required some parsers and resolvers to get the job done.

I have learned that ESLint is quite selective with what you can and cannot tell it to do (especially with typescript files).

&nbsp;

### Final Thoughts

Any of this can be modified in the event that our needs change or if @0xModene prefers some other sorting convention other than the one I have here. Just let me know so I can tweak it. Additionally, since we use Prettier for all other linting needs, I have preemptively disabled any rogue rules that were in conflict with those configs. You can browse those on line 62 in the `eslintrc.json` file. Other than that, it should work like a charm! Happy Linting 🎉

&nbsp;

### **Fixes: Issue #389**

&nbsp;

### VSC recognizes imports are out of order
eslint warnings

### Command line recognizes all imports out of order with `yarn eslint ./src`
command line warnings

### Running `yarn start` compiles with warnings
compile with warnings

### After running the `--fix` command
eslint autofix

### Successful compile after running `--fix`
successful compile

&nbsp;

###### _By submitting this pull request, you are confirming the following to be true:_


0xModene:
holy wow this is a big one. Will get to digging!

0xModene (changes requested):
This looks great, but while you're at it let's make something a bit easier. 

In `package.json:51` [here](https://github.com/SetProtocol/index-ui/blob/9afa5d4c8bc52e7e5b4c19d2a4e18b2c5daea973/package.json#L51) go ahead and make this line match what's below: 

`"lint:fix": "yarn prettier --write . && yarn eslint ./src --fix"`

This way we can have an easily accessible command to run whenever this needs to be done

0xModene:
Also noting that for whatever reason cypress isn't running here, it runs perfectly fine locally so I'm going to move forward with that in mind

0xModene (approved):
lgtm

.eslintrc.json
```diff
@@ -0,0 +1,114 @@
+{
+
+    "env": {
+        "browser": true, // // Browser global variables.
```
rootulp, line 4: Out of curiosity, why two sets of `//`? I was under the impression that comments are not supported in JSON ([ref](https://stackoverflow.com/questions/244777/can-comments-be-used-in-json)) so it's kinda cool to see that this works.

One alternative to comments in JSON is to convert this .eslintrc.json file to .eslintrc.js

rootulp (commented):
Thanks for pushing for this @tytarman ! I expect this will save us time on PR nits for import sorting

.eslintrc.json
tytarman: No problem!

The eslintrc files that are generated by `yarn eslint --init` actually allow comments in `.json` file types. I thought the same thing until I saw a ton of examples where comments were allowed. I have a `.js` file that I tested as well, and it may be good for future use because you can include functions that offer additional parameters for linting needs. 

Also, I doubled those `//` out of precaution while developing, but I guess they could be removed now. Those comments were mainly present to explain to everyone what each piece of the linter was doing because it had a lot of parts to it.

rootulp: Thanks for explaining!

Totally agreed the comments are helpful and we should preserve them. I'm okay with removing the duplicate `//` but obviously not a priority.

